PR #3, opened 2021-08-30
Add some possible changes to discuss

SamMorrowDrums:
I'll add some comments on this PR, but it can serve as a review for the initial commits.


.vscode/settings.json
```diff
@@ -6,5 +6,6 @@
 
     // transpiled JavaScript
     "lib": true,
-  }
+  },
+  "editor.formatOnSave": true,
```
SamMorrowDrums, line 10: Get Prettier to run on save.

src/codeql.test.ts
```diff
@@ -2,7 +2,7 @@ import fs from "fs";
 import path from "path";
 
 import { exec } from "@actions/exec";
-import * as io from "@actions/io";
+import io from "@actions/io";
```
SamMorrowDrums, line 5: Doing this everywhere. Not sure why `import *` ever makes sense.

src/codeql.ts
```diff
@@ -1,8 +1,10 @@
 import fs from "fs";
 import path from "path";
 
-import * as exec from "@actions/exec";
-import * as tc from "@actions/tool-cache";
+import exec from "@actions/exec";
+import tc from "@actions/tool-cache";
+
+class DatabaseUnpackingError extends Error {}
```
SamMorrowDrums, line 7: In a way this is a trivial case, and probably fine as a built-in `Error` but I like to suggest throwing custom error types as a matter of course, so that future error handling is easier to extend and to allow for handling of specific exceptions without reading their message content.

src/codeql.ts
```diff
@@ -49,29 +51,33 @@ libraryPathDependencies: codeql-${language}`
...
+  const execJSON = exec.exec(codeql, [
     "bqrs",
     "decode",
     "--format=json",
     `--output=${json}`,
     "--entities=all",
     bqrs,
   ]);
+
+  await Promise.all([execBqrs, execCSV, execJSON]);
```
SamMorrowDrums, line 79: I captured all these promises and then await them at once, so that they are not all blocking each other from starting while awaiting the completion of the previous. If you are worried about the machine's ability to handle these concurrently then perhaps this should be reverted, but generally if you await every async item individually you lose some of the benefits of async.

src/combine-results.ts
```diff
@@ -1,11 +1,26 @@
...
-import * as exec from "@actions/exec";
-import * as github from "@actions/github";
-import * as io from "@actions/io";
+import artifact from "@actions/artifact";
+import core from "@actions/core";
+import exec from "@actions/exec";
+import github from "@actions/github";
+import io from "@actions/io";
+
+const formatBody = (query: string) => `# Query
```
SamMorrowDrums, line 10: I just thought this was cleaner, as I kind of hate template type strings in function bodies - as it interrupts reading the logic.

src/interpret.ts
```diff
@@ -49,34 +48,29 @@ function toMd(tuple: any[], nwo?: string, src?: string, ref?: string): string {
...
-async function interpret(
+function interpret(
   output: stream.Writable,
   results: any,
   nwo: string,
   src: string,
   ref?: string
 ) {
-  await write(output, `## ${nwo}\n\n`);
+  output.once("open", function () {
```
SamMorrowDrums, line 60: This was the part I was most curious about. I cannot see much value in awaiting each `output.write` to flush before adding the next. I've refactored it to just write to the file async while we keep adding to it, and then end the stream and return the promisified finish event.

Also, this function is now no longer async as it simply creates the event listeners and then returns a promise, so the async function that calls it should still successfully await the finish of the write stream.

SamMorrowDrums (commented):
Comments for discussion

src/interpret.ts
SamMorrowDrums: I realise I was being silly here. What you did makes sense. It's just manually writing to streams in node has a slightly gross API.

src/codeql.test.ts
SamMorrowDrums: I realised later that this was because the actions packages don't provide a `default` export, and so TBH I prefer the style of explicit imports in that case.

src/codeql.ts
SamMorrowDrums: OK, seems that `codeql` process didn't like that. Took me a moment to realise that project didn't auto-build TS before running tests, else I'd have noticed sooner.

src/codeql.ts
```diff
@@ -49,31 +51,34 @@ libraryPathDependencies: codeql-${language}`
...
-  await exec.exec(codeql, [
+  await exec(codeql, [
     "query",
     "run",
     `--database=${database}`,
     `--output=${bqrs}`,
     queryFile,
   ]);
 
-  await exec.exec(codeql, [
```
SamMorrowDrums, line 60: I think these depend on (or cannot run at same time as) the previous exec, but I was at least able to allow these to run in parallel.

src/combine-results.ts
```diff
@@ -1,92 +1,95 @@
...
-
-\`\`\`ql
-${query}
-\`\`\`
-</details>
-
-# Results
+    const csvs: string[] = [];
+    const results = await Promise.all(
+      downloadResponse.map(async function (response) {
```
SamMorrowDrums, line 49: Given there is a lot of async stuff here that doesn't depend on each other, simply collecting the results in order as resolved promises, doing an async map makes a lot of sense.

If the order of `csvs` also matters, then we'd need to perhaps return an object and collect those values in order as well - and then iterate the response - but assuming that this function has a lot of stuff that will take a while async, it seems like a good opportunity to do it async.

I make the assumption that `csvs` doesn't need to be ordered, and that the issue comments don't need to be guaranteed to be in order either, but obviously those assumptions needs to be confirmed before this could be accepted.

src/combine-results.ts
```diff
@@ -1,92 +1,95 @@
...
-    const csvs: string[] = [];
-    for (const response of downloadResponse) {
-      const csv = path.join(response.downloadPath, "results.csv");
-      const csvDest = path.join("results", response.artifactName);
-      await io.mv(csv, csvDest);
-      csvs.push(csvDest);
-
-      const md = path.join(response.downloadPath, "results.md");
-      const comment = await octokit.rest.issues.createComment({
+    Promise.all([
```
SamMorrowDrums, line 75: Again, no need to wait for io to start other io.,

SamMorrowDrums:
@rneatherway I must say I think this all looks good without intervention, and follows conventions well. I left a mix of opinionated or alternative implementations of things that we could discuss / you could cherry pick / you could just ignore - in the spirit of sharing some thoughts and ideas I had around the code.

SamMorrowDrums:
@rneatherway in terms of TypeScript I do have one other comment. Having the `JSON.parse` of the results JSON as `any` type by default is maybe a bit of a shame. Is there a JSON schema for the results object that we could use to generate a safe interface? 

```
results["#select"]["columns"].map()
...

for (const tuple of results["#select"]["tuples"]) {
...
``` 

Both of the keys above are just assumed to exist, and would lead to unfortunate `has no properties` errors if there was to be garbage result / API change, and it might be trivial to both guard against that, and give better errors without having to write any code if we could use a type.

.vscode/settings.json
rneatherway: I was playing around with this at https://github.com/dsp-testing/qc-run2/pull/2

Don't you also need to set the default formatter? https://github.com/dsp-testing/qc-run2/commit/0bc72f40f1e5045e9d5083de75aa690cd18b2292

I was also unclear on how this would interact with eslint -- I think that _might_ be configured to run prettier as its own rules. And either way they might stomp on each other if they are both running on save.

Would you mind doing the prettier changes as a separate PR so I can try them out in isolation?

src/codeql.test.ts
rneatherway: Most of my inspiration is from https://github.com/github/codeql-action rather than any strong conviction.

I don't mind changing this kind of thing at all. Am I right in thinking that this particular change doesn't make any difference to code using the package?

src/codeql.ts
rneatherway: > ... generally if you await every async item individually you lose some of the benefits of async.

:+1: In this case I really just want to write sequential code, but the library is has an async interface, which makes sense I suppose in case you want that.

Specifically in the `codeql` case, it typically expects not to be invoked concurrently against a single project.

> Took me a moment to realise that project didn't auto-build TS before running tests, else I'd have noticed sooner.

Is there an easy way to do that? Just add it to the `test*` tasks?

src/combine-results.ts
rneatherway: Yep, definitely better!

src/codeql.test.ts
```diff
@@ -2,7 +2,7 @@ import fs from "fs";
 import path from "path";
 
 import { exec } from "@actions/exec";
-import * as io from "@actions/io";
+import { rmRF } from "@actions/io";
```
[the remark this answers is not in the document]
rneatherway: Yeah, not a fan of `import *` either. The thing that I like is that `io.rmRF` is still short and indicates to the reader which package it has come from.

src/codeql.ts
rneatherway: Ah yes, the results they want to decode won't be there yet. Nice one improving the latter two calls.

src/combine-results.ts
rneatherway: Yes, it should be safe.

rneatherway:
> Is there a JSON schema for the results object that we could use to generate a safe interface?

I don't think so, but will look into it. It's a nice idea. How would the call to `JSON.parse` then look?

rneatherway:
Successful run: https://github.com/dsp-testing/qc-controller/actions/runs/1190407113

PR #4, opened 2021-08-31
JSON typing option: Use a generated JSON type from quicktype.io

SamMorrowDrums:
One option is to throw JSON examples at: https://app.quicktype.io/ and then simply save the generated code. There are also packages that can use a JSON schema and other possibilities.

I did this one by using the JSON from `interpret.test.ts` and then generating type code from it.


rneatherway:
Thanks very nice! Please could you include a comment in the generated file indicating how to regenerate it if we need to change it?

SamMorrowDrums:
example stack trace

```
  Error {
    message: 'Invalid value undefined for type "array"',
  }

  › invalidValue (src/json-result-generated.ts:54:9)
  › transformArray (src/json-result-generated.ts:100:39)
  › transform (src/json-result-generated.ts:145:51)
  › src/json-result-generated.ts:123:30
  › Array.forEach (<anonymous>)
  › transformObject (src/json-result-generated.ts:120:41)
  › transform (src/json-result-generated.ts:146:51)
  › src/json-result-generated.ts:123:30
  › Array.forEach (<anonymous>)
  › transformObject (src/json-result-generated.ts:120:41)
```

rneatherway (approved):

PR #5, opened 2021-08-31
Add prettier configuration and update eslint config

SamMorrowDrums:
This is sufficient to auto-run prettier on save, and to ensure that prettier overrules eslint rules so you don't get any linter conflicts.

You may need to do `npm install -g eslint prettier` to ensure that plugins pick everything up but I think it works with project local versions too.

It seems we are vendoring node_modules. Normally I only save `package-lock.json`, but the vendoring conversation is bigger than this PR - so this is an aside because I'm curious.


.vscode/settings.json
```diff
@@ -6,5 +6,7 @@
 
     // transpiled JavaScript
     "lib": true,
-  }
+  },
+  "editor.formatOnSave": true,
+  "editor.formatOnPaste": false
```
rneatherway, line 11:
```suggestion
  "editor.formatOnPaste": false,
  "[typescript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
```

rneatherway: This change was required for me because of fiddling with other formatters in the past.

SamMorrowDrums: OK excellent. More concrete config is likely better anyway.

rneatherway:
No, I think this is good to go :-)

PR #7, opened 2021-09-02
Support querying multiple repositories sequentially

rneatherway:


src/codeql.ts
```diff
@@ -60,7 +60,7 @@ libraryPathDependencies: codeql-${language}`
     queryFile,
   ]);
 
-  void Promise.all([
+  await Promise.all([
```
rneatherway, line 63: @SamMorrowDrums without this change the code afterwards would sometimes run and find that these output files were not yet ready.

SamMorrowDrums: Sorry that makes a lot more sense. It was an error on my part not to `await` - the `void` was to appease the linter but the real action should have been to actually handle the response by awaiting it ✔️

src/query.ts
```diff
@@ -7,27 +11,36 @@ async function run(): Promise<void> {
...
 
     setSecret(token);
 
-    // 1. Use the GitHub API to download the database using token
-    const dbZip = await downloadDatabase(token, nwo, language);
-    await unbundleDatabase(dbZip);
+    const curDir = cwd();
+    for (const nwo of nwos) {
+      const safeNwo = nwo.replace("/", "#");
+      const workDir = mkdtempSync(path.join(curDir, safeNwo));
```
rneatherway, line 25: I initially tried using just `safeNwo`, but it was much easier to test this way because then I could query the same repository over and over. We could instead delete `workDir` after uploading the artifacts.

src/query.ts
```diff
@@ -1,3 +1,7 @@
+import { mkdirSync, mkdtempSync } from "fs";
```
SamMorrowDrums, line 1: `mkdirSync` unused.

src/codeql.ts
SamMorrowDrums: 🎉

src/query.ts
SamMorrowDrums: Doesn't seem to be any harm in this - so long as it uses the correct directory when it needs to I don't think anyone would have reason to mind. Cleaning up wastes time that an ephemeral instance shouldn't need to spend until after it's work is done.

SamMorrowDrums (approved):
This looks good - and I apologise about my mistake.

src/query.ts
SamMorrowDrums: Obviously this change is requested - so I should have said changes requested.

SamMorrowDrums (changes requested):
Will approve after the lint fix.

src/query.ts
```diff
@@ -7,27 +11,36 @@ async function run(): Promise<void> {
...
     const codeql = getInput("codeql", { required: true });
 
     setSecret(token);
 
-    // 1. Use the GitHub API to download the database using token
-    const dbZip = await downloadDatabase(token, nwo, language);
-    await unbundleDatabase(dbZip);
+    const curDir = cwd();
+    for (const nwo of nwos) {
+      const safeNwo = nwo.replace("/", "#");
```
SamMorrowDrums, line 24: Also - is that certain to be safe? Could use something like [filenamify](https://www.npmjs.com/package/filenamify) otherwise. I don't know what characters we allow - can you have emoji names for example, and can the filesystem handle them?

src/codeql.ts
robertbrignull: I agree that awaiting the promise is the right thing to do. I really like the linter check `no-floating-promises` which I guess this means you do have enabled.

There seems to be another case of using `void` to ignore the promise at
https://github.com/dsp-testing/qc-run2/blob/ac24ad467635caee2e47785a23a57180c978e694/src/combine-results.ts#L75
Although in this case we're not using the result of the promise (the worst that could happen is we say the results are available when they're not), it still looks dangerous to me to not wait for it.

Does it also stop errors from propagating? What happens if a promise that's being ignored like this fails?

src/query.ts
robertbrignull: I think the answer is that github doesn't allow any non-filename-safe characters (both linux and windows) in the repository name, or at least I couldn't use any when trying to create a repo just now. That doesn't mean there aren't repos out there with them. Using something like `filenamify` seems sensible to me too. There's also stuff like `com` to worry about, but maybe handling those is not worth it.

src/query.ts
robertbrignull: Even a self-hosted runner should clear up anything in the working directory, so I think it's fine to create whatever files/directories you want in there.

As I understand it, `mkdtempSync(path.join(curDir, safeNwo))` is creating a directory in the tmp directory, but using `path.join(curDir, safeNwo)` as a prefix and then also appending 6 random characters. So this won't ever be reusing the same directory, but the directories it creates will have very verbose names.

Is this what was intended? Have I misunderstood?

src/codeql.ts
SamMorrowDrums: @robertbrignull yes - in all honesty there is zero reason not to wait, and as mentioned that means that it will throw an error rather than that weird state of unhandled failed promise in the ether.

I don't know what I was thinking when I read the lint - but there is absolutely reason to wait for both - and even a promise that you don't need to await should at minimum have a `.error(atLeastBloodyLogIt)` even if there is no reason to hold up execution of the subsequent code.

src/query.ts
robertbrignull: No, I think I'm wrong. It does make a directory in `curDir`, using `safeNwo` as a prefix for the final directory name. So that's fine and you can ignore that part of my comment above.

And I see now that by "it was much easier to test this way because then I could query the same repository over and over" you meant so it avoids reusing the same directory so you don't have to delete it. I didn't understand initially.

src/query.ts
SamMorrowDrums: @robertbrignull on further thought, they didn't allow emoji domain names, until they did ([get an email account with one for lolz](https://mailoji.com/)) - and I think we should not take for granted that naming will remain safe - why bother having to update this in the crazy 🦄  future if we can simply handle it forever with a lib. 🦄 🦄 🦄

rneatherway: :+1: `filenamify it is!

src/query.ts
rneatherway: We might discover in the future that deletion is better because we run out of disk space, but we can cross that bridge when we come to it.

src/codeql.ts
```diff
@@ -40,6 +40,7 @@ async function runQuery(
   const bqrs = path.join("results", "results.bqrs");
   const json = path.join("results", "results.json");
   fs.mkdirSync("results");
+  fs.writeFileSync(path.join("results", "nwo.txt"), nwo);
```
robertbrignull, line 43: What's this file doing?

rneatherway:
OK, I backed out the `filenamify` change and will raise it again as a follow-up. I'll open an issue detailing the problems.

SamMorrowDrums (approved):
Looks good to me 🎉

PR #9, opened 2021-09-06
Update ava to workwith es modules

SamMorrowDrums:
Experimenting a bit more with the filenamify es modules issue.


rneatherway:
Thanks for giving this a go, but it looks like we've been able to avoid the transition for now.

PR #10, opened 2021-09-07
Fix linkification for Windows results sets

rneatherway:
Before: https://github.com/dsp-testing/qc-controller/issues/121
After: https://github.com/dsp-testing/qc-controller/issues/122


src/interpret.test.ts
```diff
@@ -40,15 +40,24 @@ const results = Convert.toJSONResult(`{
     }
   }`);
 
+const windowsResults = Convert.toJSONResult(`{
+  "#select":{"columns":[
```
robertbrignull, line 44: Not a huge deal because this is test input, but is this crazy indenting / comma positioning intentional?

robertbrignull (approved):
Changes look ok to me, and also seem to work just fine 👍

src/interpret.test.ts
rneatherway: No it wasn't! I've tidied it up now, thanks.

PR #11, opened 2021-09-07
Support signed auth tokens

robertbrignull:
Adds support for signed auth tokens + multiple tokens + using the repository id for downloading the database. I realise this is quite a lot but it perhaps makes sense to do all of this together since they're all breaking changes in some way or another.

Designed to go along with https://github.com/dsp-testing/qc-controller/pull/124 and https://github.com/github/github/pull/190233


rneatherway (approved):
LGTM

PR #12, opened 2021-09-09
Try using filenamify 4.3

rneatherway:
Rather than trying to integrate this at the same time as moving to ES modules, I'll introduce it at 4.3


SamMorrowDrums (approved):
Looks good to me. I was curious what the `nwo.txt` gets used for - but this is clearly much better than trying to change all the build to mjs in a rush just to escape filename chars.

rneatherway:
Closing in favour of https://github.com/dsp-testing/qc-run2/pull/14

PR #13, opened 2021-09-13
Remove JSON results validation for now

rneatherway:
This is causing problems with path-problem and other query types. We
can reintroduce it if we have a better schema for this but for now
it's safer to be very permissive.


robertbrignull (approved):
LGTM assuming everything still works

I'll be honest I don't think I spotted `src/json-result-generated.ts` being introduced.

PR #14, opened 2021-09-13
Use repository id for filenames

rneatherway:
Alternative that doesn't require filenamify at all. Example run: https://github.com/dsp-testing/qc-controller/actions/runs/1229921061


robertbrignull (approved):
👍  to just side stepping the dependency problem

SamMorrowDrums:
Yeah, frankly sidestepping is far simpler and this is a very low maintenance solution 👍

PR #16, opened 2021-09-15
Add integration testing workflow

rneatherway:


src/codeql.ts
```diff
@@ -81,13 +81,17 @@ libraryPathDependencies: codeql-${language}`
 }
 
 async function downloadDatabase(
-  token: string,
+  signedAuthToken: string,
   repoId: number,
-  language: string
+  language: string,
+  pat?: string
```
robertbrignull, line 87: I find it odd that `signedAuthToken ` and `pat` are not equal citizens of this argument list. Personally I'd put them next to each other and make them both nullable / undefinable.

I realise to us, using a pat is the weird case that only happens in the integration test, but in the code it's just as valid a case as a signed auth token. That's my reasoning anyway.

src/codeql.ts
```diff
@@ -81,13 +81,17 @@ libraryPathDependencies: codeql-${language}`
...
 ): Promise<string> {
+  const authHeader =
+    pat === undefined ? `RemoteAuth ${signedAuthToken}` : `token ${pat}`;
```
robertbrignull, line 90: This may need to handle the case where both `pat` and `signedAuthToken ` are null, or add a check against that.

src/query.ts
```diff
@@ -11,6 +11,7 @@ interface Repo {
   id: number;
   nwo: string;
   token: string;
```
robertbrignull, line 13: Can you make `token` and `pat` optional please. Because of the way that we're parsing the repos list with `const repos: Repo[] = JSON.parse(...)` whatever garbage object we get out of `JSON.parse` is just assumed to be a valid `Repo[]`. Any type safety only exists at compile time. So later on when we pass options to `downloadDatabase` we're actually passing an `undefined` but the code doesn't know this and happily lets us pass that as a non-optional argument.

robertbrignull (commented):
Love the integration test. I have a couple of comments about type safety and order or arguments.

src/codeql.ts
rneatherway: Sure, I had a similar feeling, I'll update it as you suggest.

src/codeql.ts
rneatherway: And in that case no auth header?

robertbrignull: I think no auth header in that case would be valid. And for a public repo could actually succeed at downloading the database.

robertbrignull (approved):
LGTM

PR #34, opened 2021-09-16
Correct off-by-one error

rneatherway:


src/combine-results.ts
```diff
@@ -59,7 +59,7 @@ async function run(): Promise<void> {
         const output = await getExecOutput("wc", ["-l", csvDest]); // TODO: preferably we would do this during results interpretation
         const results = parseInt(output.stdout.trim()) - 2;
 
-        if (results > 0) {
+        if (results >= 0) {
```
SamMorrowDrums, line 62: I'm confused - does it return `-1` for no results. The case at the bottom should would not fire anymore otherwise:

```
return `| ${repoName} | _No results_ |`;

If that's actually the case, should line 70 change to:

```
return `| ${repoName} | [${results + 1} result(s)](${comment.data.html_url}) |`;
```
```

SamMorrowDrums (commented):
Comment to clarify before I can approve.

src/combine-results.ts
```diff
@@ -59,7 +59,7 @@ async function run(): Promise<void> {
         const output = await getExecOutput("wc", ["-l", csvDest]); // TODO: preferably we would do this during results interpretation
         const results = parseInt(output.stdout.trim()) - 2;
```
SamMorrowDrums, line 60: Also, why not `const results = parseInt(output.stdout.trim()) - `;`

SamMorrowDrums:
@rneatherway also - even though it's technically not the scope of the PR, please add radix parameter to parseInt:

```
const results = parseInt(output.stdout.trim()) - 2;
```



```
const results = parseInt(output.stdout.trim(), 10) - 2;
```

https://www.oreilly.com/library/view/javascript-the-good/9780596517748/apas07.html

SamMorrowDrums:
It's true that since ECMAScript 5 , browsers treat '0101' as decimal - not octal, but it's still safest just to supply the radix - as the function still parses `0x...` as octal, and we would never want that in this case anyway.

src/combine-results.ts
rneatherway: This is the correct change! Thanks. It's a translated version of https://github.com/dsp-testing/qc-controller/blob/767b54fb9d8f876cc411e98330392ddbe4ca376c/.github/workflows/codeql-query.yml#L104

So I'm asking how many lines there are in `csvDest` excluding the header. `tail -n +2` starts counting from the second line, which is obviously not the same as subtracting 2 :facepalm:

rneatherway:
Updated :smile:

SamMorrowDrums (approved):
🎉

PR #56, opened 2021-09-24
Add support for displaying problem queries

robertbrignull:
Adds support for displaying problem queries in a nicer way.

Example: https://github.com/dsp-testing/qc-run2/issues/55

This doesn't currently handle path-problem queries at all. They will still display in raw table mode. It should be easy enough to reuse the components here to also support them but I wanted to limit the scope.

This also renames and reworks a few of the methods from `interpret.ts` as I was finding them hard to understand and hard to reuse. Hope that's ok. If you disagree with any changes there just let me know.

@rneatherway, @shati-patel


shati-patel (approved):
Thanks for doing this (and for walking me through the changes!)

I'm not 100% familiar with the code details themselves, but this seems to do the right thing 😎  This isn't particularly high-risk, so I'm happy for you to merge, and we can monitor any weirdness if it happens!

PR #65, opened 2021-09-30
Add callable workflow

rneatherway:
This moves all the remaining logic into this repository where it can
be more easily iterated on.

cc @shati-patel


.github/workflows/codeql-query.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    steps:
+      # Use ::add::mask for all tokens.
+      # Theoretically we'll only output a subset of the tokens. Unless it causes a
+      # performance issue, we may as well mask all tokens to be extra safe.
+      # Note that masking does not persist to other/future jobs in the same workflow.
+      - name: Mask tokens
+        run: |
+          jq .inputs.repositories "$GITHUB_EVENT_PATH" --raw-output | \
+            jq '[.[].token , .[].pat] | map( select( . != null ) ) | .[]' --raw-output | \
```
rneatherway, line 25: I updated this to add pat masking as well because I was testing from the command line. However, it seems like Actions may recognise the PAT format and mask them anyway because I made a mistake with my inputs and they were still filtered out: https://github.com/dsp-testing/qc-run2/runs/3755154364?check_suite_focus=true

Note: the two jq calls should be able to be folded into one using a jq pipe filter.

.github/workflows/codeql-query.yml
```diff
@@ -0,0 +1,67 @@
...
+      # and widely used. The ugly part is that is initialises a database of the
+      # given language that we just ignore.
+      - name: Initialize CodeQL
+        id: init
+        uses: github/codeql-action/init@v1
+        with:
+          languages: ${{ github.event.inputs.language }}
+
+      - name: Run query
+        uses: dsp-testing/qc-run2/query@rneatherway/usable
```
rneatherway, line 49: You'll notice that I haven't templated this part. Unfortunately that isn't possible yet, so a testing loop involving a change to the query/combine-results actions would need to temporarily update this.

.github/workflows/tmp.yml
```diff
@@ -0,0 +1,27 @@
+name: CodeQL query
```
rneatherway, line 1: This file is temporary only for testing. Successful run: https://github.com/dsp-testing/qc-run2/actions/runs/1290761996

.github/workflows/codeql-query.yml
robertbrignull: Yes, actions masks PATs or other things it recognises as a github token automatically. Still though, the explicit masking looks good to me.

.github/workflows/codeql-query.yml
```diff
@@ -0,0 +1,67 @@
...
+          codeql: ${{ steps.init.outputs.codeql-path }}
+
+  combine-results:
+    runs-on: ubuntu-latest
+    needs:
+      - run-query
+
+    steps:
+      - name: Combine results
+        uses: dsp-testing/qc-run2/combine-results@rneatherway/usable
```
robertbrignull, line 63: This branch needs changing before this is merged. I'm sure you know that, but I'm commenting anyway so it doesn't get forgotten. Also applies anywhere else that `rneatherway/usable` is mentioned.

robertbrignull (approved):
LGTM and I think this is an improvement over the current situation.

robertbrignull:
Merging now before the server-side API changes. This is just adding a file so this can be merged now.

PR #67, opened 2021-09-30
Don't fail entire actions run if some repos fail

shati-patel:
The [workflow change](https://github.com/dsp-testing/qc-run2/pull/67/files#diff-bd3f59b139152f5621e681284f72df618aac53aeda41f64130c8cbe29ec19a9e) forces the `combine-results` step to always run.
Other refinements include:
- Linking to the logs from the issue if any repos failed
- Not creating a results issue if all repos failed

I've posted some example runs in the linked issue (internal).


robertbrignull (commented):
One comment but otherwise the code looks fine. I'd be more interested to see it running. Did you say there were some example runs? Alternatively I can show you have to do a test run easily tomorrow. We might be able to trigger it using vscode which will make the testing really easy.

src/query.ts
```diff
@@ -63,8 +63,21 @@ async function run(): Promise<void> {
         { continueOnError: false, retentionDays: 1 }
       );
     }
-  } catch (error: any) {
-    setFailed(error.message);
```
robertbrignull, line 67: Do you still want to call `setFailed` (probably at the end of this `catch` block) so that the job is marked as failed? Otherwise will it be marked as a success? If a user follows the logs link we want them to be able to easily tell which jobs failed.

shati-patel: Yes, that's a good point! I will change it back.

(`setFailed` means that the entire action is marked as "failed" too, even if most of the repos succeeded. But I think that's still clearer than having the failed jobs marked as "success".)

robertbrignull: Could setting `continue-on-error: true` here help? Would need to test it but I think that would mean we could use `setFailed` to mark this step as failed but without the whole workflow being marked as a failure.

shati-patel:
Linked PR is merged, so this is good to go ⚡

PR #89, opened 2021-10-11
[WIP] Improve error messages

shati-patel:


src/download-database.ts
```diff
@@ -0,0 +1,187 @@
...
+  httpMessage: string;
+  constructor(httpStatusCode: number | undefined, httpMessage: string) {
+    super(`Unexpected HTTP response: ${httpStatusCode}. ${httpMessage}`);
+    // Set status code and error message to `this`
+    this.httpStatusCode = httpStatusCode;
+    this.httpMessage = httpMessage;
+  }
+}
+
+const userAgent = "actions/tool-cache";
```
robertbrignull, line 25: What happens if we change this to "GitHub remote queries" or something similar. Probably shouldn't pretend to be the tool cache. The user agent is pretty much free text as I understand it so can put whatever we like.

src/download-database.ts
```diff
@@ -0,0 +1,187 @@
...
+import * as httpm from "@actions/http-client";
+import { IHeaders } from "@actions/http-client/interfaces";
+import * as io from "@actions/io";
+
+export class HTTPError extends Error {
+  httpStatusCode: number | undefined;
+  httpMessage: string;
+  constructor(httpStatusCode: number | undefined, httpMessage: string) {
+    super(`Unexpected HTTP response: ${httpStatusCode}. ${httpMessage}`);
+    // Set status code and error message to `this`
```
robertbrignull, line 19:
```suggestion
```
I'm not sure this comment makes much sense and I suggest getting rid of it. If I remember rightly we just added it in the hope that copilot would do something useful. If you do want to keep it then I think it would make more sense as "set status code and error message on this" instead of "to this".

src/download-database.ts
```diff
@@ -0,0 +1,187 @@
...
+
+const userAgent = "actions/tool-cache";
+
+/**
+ * Download a database from an url and stream it into a file
+ *
+ * @param url       url of database to download
+ * @param dest      path to download database
+ * @param auth      authorization header
+ * @param headers   other headers
```
robertbrignull, line 33: We removed the `headers` arg.

src/download-database.ts
```diff
@@ -0,0 +1,187 @@
...
+  core.debug(`Downloading ${url}`);
+  core.debug(`Destination ${dest}`);
+
+  const maxAttempts = 3;
+  const minSeconds = 10;
+  const maxSeconds = 20;
+  const retryHelper = new RetryHelper(maxAttempts, minSeconds, maxSeconds);
+  return await retryHelper.execute(
+    async () => {
+      return await downloadDatabaseFileAttempt(url, dest || "", auth);
```
robertbrignull, line 51: `dest` is always defined, so `dest || ""` is unnecessary. 
```suggestion
      return await downloadDatabaseFileAttempt(url, dest, auth);
```

src/download-database.ts
shati-patel: Thanks! That worked 🎉

shati-patel:
Thanks for the review!

PR #108, opened 2021-10-20
Use query pack, if given

rneatherway:
- [x] Decide about base64/not
- [x] Add a test of the query pack mode
- [x] Test manually using workflow_dispatch
- [x] Rename `downloadDatabase` to something more generic


query/action.yml
```diff
@@ -1,22 +1,26 @@
...
+description: "Run a CodeQL query against a repository"
+author: "GitHub"
 inputs:
   query:
-    description: 'Text of CodeQL query to run'
+    description: "Text of CodeQL query to run. Exactly one of 'query' and 'query_pack_url' are required."
+    required: true
+
+  query_pack_url:
+    description: "URL for a base64 encoded tarball containing CodeQL query plus dependencies. Exactly one of 'query' and 'query_pack_url' are required."
```
robertbrignull, line 10: Just checking what exactly "URL for a base64 encoded tarball" means and is it correct? It seems the base64 encoding is unnecessary if it's going to be downloaded via a URL.

If this is just the way it is and changing it would be effort then that's fine. Just wanted to check as it seemed odd to me.

src/query.ts
```diff
@@ -4,8 +4,10 @@ import { chdir, cwd } from "process";
 
 import { create as createArtifactClient } from "@actions/artifact";
 import { getInput, setSecret, setFailed } from "@actions/core";
+import { extractTar as extractTar } from "@actions/tool-cache";
```
robertbrignull, line 7: Is this rename of the imported method doing anything?

src/query.ts
```diff
@@ -49,9 +58,19 @@ async function run(): Promise<void> {
         repo.pat
       );
 
+      let queryPack: string | undefined;
+      if (queryPackUrl !== "") {
+        console.log("Getting query pack");
+        const queryPackArchive = await downloadDatabaseFile(
```
robertbrignull, line 64: Not strictly necessary right now but I'm feeling we should rename `downloadDatabaseFile` and the file it's in to something not mentioning databases. It's just a generic method for downloading whatever.

query/action.yml
rneatherway: It's because the API call passes it in the JSON body as a base64 string. We could decode it in the API handler before inserting it into S3, but that doesn't happen currently. Do you think that we should make that change?

If not, you've reminded me that I need to do the base64 decoding here!

robertbrignull: I agree it makes sense to do the decoding here rather than in the API handler. So leaving it as it is (i.e. base64 encoded) is fine 👍

PR #154, opened 2021-11-08
Results viewing: link to correct commit instead of "HEAD"

shati-patel:
Sorry, most of these "Files changed" are node modules 😭  (I added `js-yaml` to read the `codeql-database.yml` metadata file... Maybe there's a nicer way 😕)

The actual changes:
- [codeql.ts](https://github.com/dsp-testing/qc-run2/pull/154/files#diff-473ad8cf33ad034dc4f9ae468a3ce5d87106ace4f96f6a23a000261d46efaa60): Adds a function that returns the commit SHA for the database. We then use that to link to results.
- [codeql.test.ts](https://github.com/dsp-testing/qc-run2/pull/154/files#diff-a59b656ff6cb8e9e4da88f2f927636ffc66ffadca08a86eb1024f9c2690d8ad9): Tests `getDatabaseSHA` with a fake new database that contains a commit SHA in its metadata, and with a database built with the current CLI bundle (which doesn't have a commit SHA)
  - Note: once the bundle is updated, this test will fail and we should update the test (so that the fake db doesn't have a SHA, and the real one does).

Could you check that these changes makes sense? I'm especially unfamiliar with the ava test library, so I roughly copied the existing tests 🙃 

(Also, this will conflict with #161, but it should hopefully be easy enough to fix...)


robertbrignull (commented):
A few comments for more error checking, but generally looks pretty much fine to me and definitely along the right path.

src/codeql.test.ts
```diff
@@ -89,3 +89,32 @@ test("running a query in a pack", async (t: any) => {
...
+test("getting the commit SHA for a database", async (t: any) => {
+  // Note: db-metadata-only.zip is not a real database.
+  // It doesn't contain source code, only a codeql-database.yml file.
+  const testDbZip = path.resolve("testdata/db-metadata-only.zip");
+  const tmpDir = fs.mkdtempSync("tmp");
+  const cwd = process.cwd();
+  process.chdir(tmpDir);
+  try {
+    // Test our fake database. This "database" has a commit SHA in its metadata.
+    await exec("codeql", ["database", "unbundle", testDbZip, "--name=fakeDb"]);
```
rneatherway, line 102: It's preferable to avoid checking in binary files where possible (although not a hard and fast rule), so you could remove this line and just have the database be a directory rather than a zip file.

src/codeql.test.ts
```diff
@@ -89,3 +89,32 @@ test("running a query in a pack", async (t: any) => {
...
+    const sha1 = getDatabaseSHA("fakeDb");
+    t.is(sha1, "ccf1e13626d97b009b4da78f719f028d9f7cdf80");
+
+    // Test the real database, which has no commit SHA (until CodeQL CLI 2.7.2 is released).
+    // TODO: update this test once CodeQL CLI 2.7.2 is released.
```
rneatherway, line 107: This is an interesting case because the test will suddenly start failing on `main` without us having taken any action. On a larger project this could really confuse other people, but here where all three of us are aware I think it's completely fine, and actually useful for us to know when the feature will start working.

rneatherway:
> Sorry, most of these "Files changed" are node modules 😭 (I added `js-yaml` to read the `codeql-database.yml` metadata file... Maybe there's a nicer way 😕)

I think that's just the way it is, don't worry about it :-)

shati-patel:
Thanks for the reviews (and @robertbrignull for helping me make changes 😅)
I ended up removing the test database altogether and writing the metadata YAML to a temporary file instead.

robertbrignull (approved):
LGTM

I looked in the raw test logs too and it seems our error checking is working and the test data is creating the desired errors:
```
2021-11-09T12:11:33.2959356Z Unable to read codeql-database.yml: YAMLException: end of the stream or a document separator is expected (2:1)
2021-11-09T12:11:33.2959940Z 
2021-11-09T12:11:33.2960207Z  1 |     foo:"
2021-11-09T12:11:33.2960458Z  2 | bar
2021-11-09T12:11:33.2960841Z -----^
2021-11-09T12:11:33.3003860Z   ✔ codeql.ts › getting the commit SHA when codeql-database.yml exists, but is invalid
2021-11-09T12:11:33.3012394Z Unable to read codeql-database.yml: Error: ENOENT: no such file or directory, open 'tmpW7VtNv/codeql-database.yml'
2021-11-09T12:11:33.3053232Z   ✔ codeql.ts › getting the commit SHA when the codeql-database.yml does not exist
```

PR #161, opened 2021-11-08
Output sarif file for problem queries

robertbrignull:
This PR makes it so that for problem queries we output and upload a SARIF file.

Example run on a non-problem query: https://github.com/dsp-testing/qc-controller/actions/runs/1435448966
Example run on a problem query: https://github.com/dsp-testing/qc-controller/actions/runs/1435449526

We now use `codeql bqrs info` to determine if the query is a problem query, as this gives a very handy list of formats that the query adheres to.

This also changes around the function definitions in `codeql.ts` to run more of them in parallel, and changes how we specify which files to upload because this is now variable and we need to specify the correct set.


src/codeql.ts
```diff
@@ -131,3 +110,93 @@ async function downloadDatabase(
...
+
+// Generates results.sarif from the given bqrs file
+async function outputSarif(codeql: string, bqrs: string): Promise<string> {
+  const sarif = path.join("results", "results.sarif");
+  await exec(codeql, [
+    "bqrs",
+    "interpret",
+    "--format=sarif-latest",
+    `--output=${sarif}`,
+    "-t=kind=problem",
```
robertbrignull, line 197: These two options were required by `codeql bqrs interpret`. I think we have to just make something up for these, because the query might not have them, and even if it did then getting them from the query text could be difficult. I don't think the values matter particularly right now. Maybe in the far future if there's some option to import the results into code scanning it could be important, but that's not a realistic possibility for now.

src/codeql.ts
```diff
@@ -131,3 +110,93 @@ async function downloadDatabase(
...
+    encoding: "utf8",
+  });
+
+  await interpret(
+    s,
+    jsonResults,
+    nwo,
+    compatibleQueryKinds,
+    sourceLocationPrefix,
+    "HEAD"
```
rneatherway, line 184: This looks like it might conflict with #154

src/codeql.ts
rneatherway: Couldn't this argument's value come from the output of `codeql bqrs info`? This path should ideally work for both `problem` and `path-problem`.

src/interpret.ts
```diff
@@ -123,7 +124,7 @@ async function interpret(
 
   await write(output, `## ${nwo}\n\n`);
 
-  if (isProblemQuery(results)) {
+  if (compatibleQueryKinds.includes("Problem")) {
```
rneatherway, line 127: Very nice improvement

src/query.ts
```diff
@@ -70,18 +70,20 @@ async function run(): Promise<void> {
 
       // 2. Run the query
       console.log("Running query");
-      await runQuery(codeql, language, dbZip, repo.nwo, query, queryPack);
+      const filesToUpload = await runQuery(
```
rneatherway, line 73: Thanks 👍

src/codeql.ts
robertbrignull: For info, here's what the `codeql bqrs info --format=json` output looks like
```
{
  "result-sets" : [
    {
      "name" : "#select",
      "rows" : 0,
      "columns" : [
        {
          "name" : "getInputCall",
          "kind" : "e"
        },
        {
          "kind" : "s"
        },
        {
          "name" : "inputName",
          "kind" : "s"
        },
        {
          "name" : "inputName",
          "kind" : "s"
        },
        {
          "name" : "action",
          "kind" : "e"
        },
        {
          "kind" : "s"
        }
      ]
    }
  ],
  "compatible-query-kinds" : [
    "Problem",
    "Table",
    "Tree",
    "Graph",
    "Diagnostic"
  ]
}
```
So the id is completely gone, and the kind is determinable from the compatible query kinds. So you're right we do want this to work for path problems too.

robertbrignull:
Here are some more example runs:
- problem query without any metadata: https://github.com/dsp-testing/qc-controller/actions/runs/1439411369
- problem query with metadata: https://github.com/dsp-testing/qc-controller/actions/runs/1439413739
- path-problem query: https://github.com/dsp-testing/qc-controller/actions/runs/1439416192
- non-problem query: https://github.com/dsp-testing/qc-controller/actions/runs/1439428152

src/codeql.ts
robertbrignull: Yes, but hopefully won't be too hard to resolve. @shati-patel, just let me know if it's hard to resolve.

shati-patel: Should be fine! Thanks for the heads-up 🌷

robertbrignull:
The conflicts turned out to be nastier than I thought they'd be, but I think I've fixed them all correctly.

New test run: https://github.com/dsp-testing/qc-controller/actions/runs/1439491154

robertbrignull:
@rneatherway, should be ready for re-review now

src/codeql.ts
rneatherway: :+1: A fake id should be fine AFAICT

rneatherway (approved):

PR #188, opened 2021-11-09
Upload a result index file

shati-patel:
Updates the `combine-results` action to upload an index file with general info about the results:
```
[
  {
    "nwo": ...,
    "id": "...",
    "results_count": ...,
    "bqrs_file_size": ...,
    "sarif_file_size": ... 
  }, 
  {
    ...
  }
]
```
(`sarif_file_size` only exists for problem and path-problem queries)

To make the refactoring a bit easier, I've also fixed a "TODO" about counting the number of results during interpretation. 

--- 

(🍐 paired with @robertbrignull on this!)


robertbrignull (approved):
LGTM 👍

src/combine-results.ts
```diff
@@ -94,34 +93,71 @@ async function run(): Promise<void> {
...
+        return csvDest;
+      })
+    );
+    const resultsMd = await Promise.all(
+      resultArtifacts.map(async function (response) {
         const repoName = fs.readFileSync(
           path.join(response.downloadPath, "nwo.txt"),
           "utf-8"
         );
-        const output = await getExecOutput("wc", ["-l", csvDest]); // TODO: preferably we would do this during results interpretation
```
rneatherway, line 109: Thanks!

src/combine-results.ts
```diff
@@ -94,34 +93,71 @@ async function run(): Promise<void> {
...
             body: fs.readFileSync(md, "utf8"),
           });
-          return `| ${repoName} | [${results} result(s)](${comment.data.html_url}) |`;
+          return `| ${repoName} | [${resultCount} result(s)](${comment.data.html_url}) |`;
         }
         return `| ${repoName} | _No results_ |`;
       })
     );
+    const resultsIndex = await Promise.all(
+      resultArtifacts.map(async function (response) {
```
rneatherway, line 132: Again, not critical, but a custom type to build and then serialize would be useful internal documentation.

I also think there is an interesting test opportunity here where you factor out this part into another file (you can't test against the "main" action file because it tries to run it!) and then call `runQuery` followed by this logic pointed out the output directory.

rneatherway (commented):
Two comments, both non-blocking

shati-patel:
Okay, I think the test finally works! 😭 😅 
@rneatherway, let me know if you have more suggestions for the general structure or for the `BQRSInfo` type 🙇🏽‍♀️

src/combine-results.ts
```diff
@@ -94,34 +94,46 @@ async function run(): Promise<void> {
...
             issue_number: issue.data.number,
             body: fs.readFileSync(md, "utf8"),
           });
-          return `| ${repoName} | [${results} result(s)](${comment.data.html_url}) |`;
+          return `| ${repoName} | [${resultCount} result(s)](${comment.data.html_url}) |`;
         }
         return `| ${repoName} | _No results_ |`;
       })
     );
+    const resultsIndex = createResultIndex(resultArtifacts);
```
rneatherway, line 132:
```suggestion
    const resultsIndex = await createResultIndex(resultArtifacts);
```

Perhaps `JSON.stringify` waits internally somehow, but this feels more "correct"

src/codeql.ts
```diff
@@ -118,11 +120,16 @@ async function downloadDatabase(
...
-): Promise<string[]> {
+interface BQRSInfo {
+  "result-sets": Array<{
+    name: string;
+    rows: number;
+  }>;
+  "compatible-query-kinds": string[];
+}
+
+// Calls `bqrs info` for the given bqrs file and returns JSON output
```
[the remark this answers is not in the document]
rneatherway: We agreed to defer this.

rneatherway:
A single small comment

src/combine-results.ts
robertbrignull: No, I think we were stringifying a promise here, so we do need to `await`. Another thing you could do is explicitly add a type to `resultsIndex` to assert it is what we want it to be, e.g.
```suggestion
    const resultsIndex: ResultIndexItem[] = await createResultIndex(resultArtifacts);
```

shati-patel: Thank you, that makes sense!

rneatherway: Ah! I think the integration test "caught" this (the results-index/index.json at https://github.com/dsp-testing/qc-run2/actions/runs/1449152312 is just `{}`). That integration test just checks that we aren't crashing though (which is fine).

PR #193, opened 2021-11-10
Add scripts and workflow for testing branches

rneatherway:
VSCode seems to be formatting YAML for me, so I went with it.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -7,49 +7,58 @@ on:
   workflow_dispatch:
 
 jobs:
+  workflow-references-main:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check workflow references only main
+        run: |
+          ! grep 'uses: dsp-testing/qc-run2/' .github/workflows/codeql-query.yml | grep -q -v @main
+
   lint-js:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Run Lint
-      run: npm run-script lint
+      - uses: actions/checkout@v2
+      - name: Run Lint
+        run: npm run-script lint
 
   check-js:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Check generated JavaScript
-      run: .github/workflows/script/check-js.sh
+      - uses: actions/checkout@v2
+      - name: Check generated JavaScript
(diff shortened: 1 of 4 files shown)


rneatherway:
@shati-patel @robertbrignull what do you think of these scripts and test for facilitating switching out the branches for testing?

.github/workflows/pr-checks.yml
```diff
@@ -7,49 +7,58 @@ on:
...
 
 jobs:
+  workflow-references-main:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check workflow references only main
+        run: |
+          ! grep 'uses: dsp-testing/qc-run2/' .github/workflows/codeql-query.yml | grep -q -v @main
```
robertbrignull, line 17: I tested it and indeed it is exit code 0 iff all references are using `@main` and it's exit code 1 if any of the references are some other branch. So it seems to work.

But, just to make sure I understand how this bash is working:
- `grep 'uses: dsp-testing/qc-run2/' .github/workflows/codeql-query.yml` outputs all lines containing `uses: dsp-testing/qc-run2/`. Nice and simple. It will be exit code 0 iff there is at least one line matching that, but that should always be the case so that's fine.
- `grep -q -v @main` is in quiet mode and is searching for lines that do not contain `@main`. So the output to stdout is suppressed, and it'll be exit code 0 iff at least one line doesn't contain `@main`.
- We then use `!` to invert the success/failure and I'm assuming this applies to the result of the whole line including both greps and the pipe, rather than just the first grep. So now the whole thing will be exit code 1 iff at least one line contains `uses: dsp-testing/qc-run2/` but not `@main`.

So I guess there's a slight edge case for a branch like `@main2` but I think we can fix that with
```suggestion
          ! grep 'uses: dsp-testing/qc-run2/' .github/workflows/codeql-query.yml | grep -q -v '@main$'
```

robertbrignull (approved):
LGTM with one tiny suggested improvement.

It would be nice to also mention that these scripts exist in a the readme.

shati-patel:
> @shati-patel @robertbrignull what do you think of these scripts and test for facilitating switching out the branches for testing?

This is really useful! Thank you 🎉  
(I don't have anything to add to @robertbrignull's comments above)

.github/workflows/pr-checks.yml
rneatherway: I agree, and thanks for the bugfix.

PR #214, opened 2021-11-11
Try camelCasing

rneatherway:
The test I added is more of an integration test, which should make sure that we check against the current version of CodeQL. It feels a bit like overkill, but we were running a query anyway so we have the `.bqrs` to hand.

How does this look?


robertbrignull (approved):
LGTM

PR #236, opened 2021-11-12
Address problems with issues being too large, or hitting the rate limit when posting

robertbrignull:
Truncates issue comment bodies when they become too large. Checks that each row of the results table will fit before trying to add it. This is the bit with the generators and I hope it's not too complicated.

Also adds a tiny delay between posting issue comments to avoid hitting the rate limit. To do this the `Promise.all` had to be changed to a `for` loop so we would process each repo in series, otherwise putting a sleep in the async method would do nothing.

Currently untested. I'll try to add some unit tests and do some manual testing.


robertbrignull:
I've made some changes, fixed the bugs we spotted in the meeting, and updated the messages to be more informative.

Here's what it looks like when an individual comment is truncated: https://github.com/dsp-testing/qc-controller/issues/290
And when the number of repositories is too large: https://github.com/dsp-testing/qc-controller/issues/291

src/interpret.ts
```diff
@@ -135,33 +139,183 @@ async function interpret(
...
+// Outputs a table to the writable stream.
+// Avoids going over the issue comment length limit and will truncate results if necessary.
+// Returns the number of rows written to the output, excluding any header rows.
+async function writeTableContents(
+  output: stream.Writable,
+  colNames: string[],
+  nextRow: Generator<string, undefined, unknown>
+): Promise<number> {
+  // Issue comment limit is 65536 characters.
+  // But eave a bit of buffer to account for the comment title and truncation warning text.
```
rneatherway, line 199:
```suggestion
  // But leave a bit of buffer to account for the comment title and truncation warning text.
```

rneatherway (approved):

PR #239, opened 2021-11-15
Change repo/nwo/token input to a download

rneatherway:
This input exceeds the Actions size limits around 500 repositories,
depending on nwo length.

I've kept the flow as similar as possible to before to limit the size
of the change, so the workflow still expects the repositories to be
chunked before being called. With this better testing in place we
could perhaps change that later, although I don't see a strong benefit
right now.

The two non-constant inputs are moved to a JSON file at
`inputs.instructions_url` with format:

```json
{
  "repositories": [
    {
      "id": 10,
      "nwo": "a/b",
      "token": "..."
    },
    {
      "id": 12,
      "nwo": "b/c",
      "token": "..."
    }
  ]
  "repoNwoChunks": [
    ["a/b"],
    ["b/c"]
  ]
}
```

We are then very careful to mask the download URL throughout the
workflow.

For testing, the instructions json is placed in the test release along
with the query pack, so it isn't secret and therefore has to elide the
tokens. I've stuck with using a PAT stored as a repository secret for
that. As we cannot inject it directly to an input as before I have
instead added a special case to the workflow to inject the PAT if the
secret input `TEST_PAT` is present.

I've also updated the integration tests with a new one that uses the
entire callable workflow (`codeql-query.yml`). This test does have the
downside that again the ref to run has to be hard-coded, so
`script/use-*` have been taught to also change that test. I plan to
later remove the other integration tests despite this limitation, but
that's up for discussion. To mimic the way we use the inputs from the
webhook file I had to trigger it using workflow_dispatch with the
correct inputs.

Test: workflow_dispatch -> workflow_call    -> codeql_query.yml
Real: github            -> dynamic workflow -> codeql_query.yml


robertbrignull (approved):
LGTM if it works when tested. Well done for finding a way to handle the secrets during tests.

PR #343, opened 2021-11-16
Use perl rather than sed

rneatherway:


robertbrignull (approved):
WFM

Thanks for doing an alternative

rneatherway:
np, I have a soft spot for perl

PR #382, opened 2021-11-23
Escape markdown from query results

rneatherway:
This one is fiddly to test against the exact example because that one uses a query pack, but I've added unit tests and you can see the general idea at https://github.com/dsp-testing/qc-controller/issues/299.


src/interpret.ts
```diff
@@ -62,15 +63,33 @@ function getEntityURL(e: any, nwo: string, src: string, ref: string): string {
...
 // Returns the formatted message for a problem query, with any placeholders filled in.
 function problemQueryMessage(
   tuple: any,
   nwo: string,
   src: string,
   ref: string
 ): string {
-  // Start with just the raw message, and then fill in any placeholders
-  let message = tuple[1] as string;
+  // Start with just the raw message, and then fill in any placeholders We
```
robertbrignull, line 87:
```suggestion
  // Start with just the raw message, and then fill in any placeholders. We
```

robertbrignull (approved):
One trivial comment, otherwise LGTM

src/interpret.ts
rneatherway: Thanks

PR #392, opened 2021-11-24
Do not override default artifact retention

rneatherway:


shati-patel (approved):
Super speedy! 

Looks good 🌷

PR #399, opened 2021-11-29
Avoid using sync methods during parallel sections

robertbrignull:
The aim of this PR is to remove any places where we do a synchronous IO operation in a place that it designed to run in parallel with other code. If we change `fs.readFileSync` to `await fs.promises.readFile` then it hopefully will avoid blocking any other threads and instead let them progress while the file is being read. This probably won't make any noticeable difference in practice but it's nice. I realise too this is a fair bit of churn and not the highest importance.

For `src/codeql.ts` all that was needed was changing two lines in the above way.

For `src/combine-results.ts` I went slightly further and pulled code out to methods. There are essentially three things we're doing there (updating the issue body, uploading CSV files, uploading result index file) and we were doing them one after the other when we could do them all in parallel. Hopefully splitting it up also makes it easier to read and see what depends on what.

It's sad that you have to do `await fs.promises.readFile`. Worth noting we could switch to using `fs-extra` instead which makes the interface a bit nicer, so we can just do `await fs.readFile`, and it adds other nice methods. But I didn't want to add another dependency just for this.

@rneatherway


rneatherway (approved):
If the tests pass it's fine by me :smile:

PR #406, opened 2021-12-01
Execute the default suite from a query pack

rneatherway:
This now works with the latest version of the VSCode changes. Example: https://github.com/dsp-testing/qc-controller/actions/runs/1526840822

I'm happy with it apart from the fairly ad-hoc parsing of the `qlpack.yml` to find the query that was executed for inclusion in the markdown, but that is a best-effort display anyway really.


src/codeql.ts
```diff
@@ -65,13 +72,46 @@ libraryPathDependencies: codeql-${language}`
...
+      `--database=db`,
+      `--output=${bqrs}`,
+      queryFile,
+    ]);
+  } else if (queryPack !== undefined) {
+    await exec(codeql, [
+      "database",
+      "run-queries",
+      "--additional-packs",
+      queryPack,
```
aeisenberg, line 88: This is the path to the directory containing the `queryPack`? If so, this is fine.

src/codeql.ts
```diff
@@ -280,3 +320,39 @@ function getDatabaseSHA(database: string): string {
...
+      fs.readFileSync(path.join(queryPack, "qlpack.yml"), "utf8")
+    ) as QLPack | undefined;
+  } catch (error) {
+    console.log(`Unable to read qlpack.yml: ${error}`);
+    return;
+  }
+
+  if (metadata?.defaultSuite) {
+    const entry = metadata.defaultSuite.find((elem) => elem.query);
+    if (entry && entry.query) {
```
aeisenberg, line 349: nit:

```suggestion
    if (entry?.query) {
```

aeisenberg (commented):
This looks entirely reasonable to me. It's unfortunate that you need to parse the qlpack.yml file to get the query you just ran, but I can't think of a better way off the top of my head. This is brittle in case we ever start running multiple queries. Maybe instead of runnng `metadata.defaultSuite.find`, you can run `reduce` and collect all queries specified in the suite? This is a detail that can be changed later.

src/codeql.ts
rneatherway: Yes, that's right.

src/codeql.ts
rneatherway: Thanks!

rneatherway:
> This looks entirely reasonable to me. It's unfortunate that you need to parse the qlpack.yml file to get the query you just ran, but I can't think of a better way off the top of my head. This is brittle in case we ever start running multiple queries. Maybe instead of runnng `metadata.defaultSuite.find`, you can run `reduce` and collect all queries specified in the suite? This is a detail that can be changed later.

Yeah, I agree. Right now though I think there is still a chance that this markdown output will be phased out once we can easily view the results in VSCode, so if you don't mind I'll leave it like this.

src/codeql.ts
```diff
@@ -280,3 +320,31 @@ function getDatabaseSHA(database: string): string {
...
+ * @param queryPack The path to the query pack on disk.
+ * @returns The path to a query file.
+ */
+async function getRemoteQueryPackDefaultQuery(
+  codeql: string,
+  queryPack: string
+): Promise<string> {
+  const output = (
+    await getExecOutput(codeql, [
+      "resolve",
```
aibaars, line 337: You could use `--format=json` which returns a JSON array of strings. Parsing JSON might be more robust than splitting by `\n` . It's unlikely the CLI would ever produce and old-style MacOS line ending (`\r` without `\n`) or a query path would contain a newline, but still.

rneatherway: Thanks, that is much better! I didn't think to check that this command supported the format option.

rneatherway:
OK tested: https://github.com/dsp-testing/qc-controller/actions/runs/1527235600. Integration tests also pass.

aeisenberg (approved):
Nice. The `resolve queries` idea was a good one.

PR #445, opened 2021-12-03
Fix interpretation of entities without locations

rneatherway:


src/interpret.test.ts
```diff
@@ -438,3 +448,34 @@ test("escaping markdown works", (t) => {
     "This part of the regular expression may cause exponential backtracking on strings starting with '&#60;table=id=\"szamlat\"&#62;a&#60;table' and containing many repetitions of '&#60;/table&#62;a&#60;table'&#46;"
   );
 });
+
+test("external api results converted correctly", async (t) => {
+  let output = "";
+  const w = new Stream.Writable({
```
charisk, line 454: Happy to be told to RTFM but why are we having to implement how a writable stream works?

charisk (approved):

src/interpret.test.ts
charisk: Ah just did some reading! Ignore me.

PR #520, opened 2022-01-04
Remove integration test that doesn't work

rneatherway:
The source root is not a git repo, so this test wouldn't have worked. My bad for suggesting it.


shati-patel (approved):
💥 Thanks! 🔪

PR #550, opened 2022-02-04
Remove unused artifact + fail action when there are no successful runs

shati-patel:
- Deletes the `all-results` artifact, which is no longer used
- Fails the action if the remote query didn't succeed on any repos

Test runs in our internal controller repo 🙂


rneatherway (approved):
Looks good. In the future I suppose we could also get rid of producing the per-repo CSV files as well: https://github.com/dsp-testing/qc-run2/blob/13c6083ea30eaf2115a1f801354406608fabcee8/src/codeql.ts#L177

PR #567, opened 2022-02-15
Add error information to result-index

shati-patel:
Do not merge yet! Requires changes to the CodeQL extension to be released (not just merged!) first: https://github.com/github/vscode-codeql/pull/1159. 

cc @rneatherway

----

Instead of having a single "errors" artifact, we now upload a "<repo.id>-error" artifact for each repository that failed. This is so the VS Code extension can read/display that information 💃🏽 See internal linked issue 🔗 

The result index now looks like this:
```
{
  "successes": [
    {
      "nwo": "dsp-testing/qc-demo-github-certstore",
      "id": "376068344",
      "results_count": 0,
      "bqrs_file_size": 23,
      "sarif_file_size": 759
    }
 ],
  "failures": [
    {
      "nwo": "dsp-testing/qc-demo-github-freno",
      "id": "376065647",
      "error": "Language mismatch: The query targets go, but the repository \"dsp-testing/qc-demo-github-freno\" has no CodeQL database available for that language."
    },
    {
      "nwo": "dsp-testing/qc-demo-github-vulcanizer",
      "id": "376069179",
      "error": "Language mismatch: The query targets go, but the repository \"dsp-testing/qc-demo-github-vulcanizer\" has no CodeQL database available for that language."
    }
  ]
}
```

Note: we also updated the integration test (i.e. the instructions.json file) to run on a "broken" repo.


src/interpret.ts
```diff
@@ -339,17 +339,26 @@ function timeout(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
-interface ResultIndexItem {
+export interface SuccessIndexItem {
```
charisk, line 342: I think it's worth adding a `type` or `t` property here (or whatever convention is used elsewhere) to explicitly call out whether something is a success or a failure.

At the minute, in the VS Code extension code, we rely on the presence of an `error` property which works but isn't great and makes our types a bit confusing.

Thoughts @rneatherway?

rneatherway: Yes, I agree. I found this in the older version of the TypeScript Handbook: https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions, which suggests the same. I'm a bit disappointed that there isn't a type-level `switch` but I guess you can't win them all.

shati-patel: Changed the shape of the result index in https://github.com/dsp-testing/qc-run2/pull/567/commits/aadf5bb7edd44fb10e1af30aa167fcbe47806d82 (and tested with corresponding changes in the VS Code extension). Seems to work 🎉

shati-patel:
Could I get another review? Then I'll revert that temp change for testing 😁

rneatherway (approved):
Nice! A good sign that the test changes are so small.

shati-patel:
Reverted the revert for now so we can continue to use this branch for testing! cc @aeisenberg

.github/workflows/codeql-query.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           CODEQL_ENABLE_EXPERIMENTAL_FEATURES: true
 
       - name: Run query
-        uses: dsp-testing/qc-run2/query@main
+        uses: dsp-testing/qc-run2/query@better-errors
```
aeisenberg:
```suggestion
        uses: dsp-testing/qc-run2/query@main
```

.github/workflows/codeql-query.yml
```diff
@@ -98,7 +98,7 @@ jobs:
           CODEQL_ENABLE_EXPERIMENTAL_FEATURES: true
 
       - name: Combine results
-        uses: dsp-testing/qc-run2/combine-results@main
+        uses: dsp-testing/qc-run2/combine-results@better-errors
```
aeisenberg:
```suggestion
        uses: dsp-testing/qc-run2/combine-results@main```

aeisenberg (commented):
Move back to using `main` in coordination with the 1.6.0 release of vscode-codeql.

aeisenberg (approved):
Thanks.

PR #612, opened 2022-02-22
Add script for easier access to `codeql` when running tests in codespaces

shati-patel:
You need `codeql` on the path to run tests. Here's an attempt to make that easier to do in Codespaces. If you run `script/test` instead of `npm run test`, everything is great.


script/add-codeql-to-path
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -eu
+
+cd "$(dirname "$0")/.."
```
shati-patel, line 5: Not sure if we needed this 🙈  I just copied these lines from the other scripts 😂

script/add-codeql-to-path
```diff
@@ -0,0 +1,11 @@
...
+
+gh extensions install github/gh-codeql
```
shati-patel, line 7: Ideally, I want to check whether `gh codeql` is already installed, and skip this step if so...

script/add-codeql-to-path
```diff
@@ -0,0 +1,12 @@
...
+gh codeql download
```
shati-patel, line 8: Ideally, I want to check whether `gh codeql` is already installed/downloaded, and skip this step if so...

script/add-codeql-to-path
rneatherway: :grin: The purpose is to get you to the root of the repository, so that it doesn't matter what directory you're in when you run the script. There are lots of ways of achieving the same thing, but this is one of them.

In Bash the arguments that you pass on the command line are stored in `$1`, $2`, etc... and `$0` is used for the path to the script itself. That'll be something like `/home/rneatherway/qc-run2/script/add-codeql-to-path`, so then `dirname` will strip it back to `/home/rneatherway/qc-run2/script` and when you add the `..` it's effectively ``/home/rneatherway/qc-run2`.

script/add-codeql-to-path
rneatherway: I just tried it and it looks like the command knows to check this so you don't need to worry about that.

script/add-codeql-to-path
```diff
@@ -0,0 +1,12 @@
...
+
+export PATH=$PWD:$PATH
```
rneatherway, line 10: This part is a bit fiddly. I'm afraid that this won't work because when you run this script from your shell it runs in a subprocess that can't affect the environment variables (including PATH) in its parent process.

We've got two options:
1. Make a `script/test` script that does this and then runs `npm test "$@"` once the PATH is correct. This is nice and simple, but doesn't help if you want `codeql` on the PATH for other reasons.
2. You can make a non-executable bash script that you "source" instead using `source script/add-codeql-to-path` or equivalently `. script/add-code-to-path`, which basically behaves the same as if you typed the commands into your shell.

script/add-codeql-to-path
```diff
@@ -0,0 +1,12 @@
...
+
+chmod +x codeql
```
rneatherway, line 12: I'd suggest putting `codeql` in the `script` directory to keep things a bit cleaner. Git will remember the `chmod +x` for us so you don't need this line.

shati-patel:
(Merging since @rneatherway helped me get this working 💚)

PR #639, opened 2022-03-01
Add database sha to result index

shati-patel:
Attempt to add database SHA to the result-index, so that the VS Code extension has access to this info! 
Tested in conjunction with https://github.com/github/vscode-codeql/pull/1205

(See internal issue)


shati-patel:
Closing in favour of SARIF `versionControlProvenance` approach.

src/codeql.ts
```diff
@@ -55,6 +55,8 @@ async function runQuery(
   console.log(
     `This database was created using CodeQL CLI version ${dbMetadata.creationMetadata?.cliVersion}`
   );
+  const databaseSHAFile = path.join("results", "sha.txt");
+  fs.writeFileSync(databaseSHAFile, dbMetadata.creationMetadata?.sha || "");
```
rneatherway, line 59: The failure case is the only thing that I'm unsure about here.

How would you expect a `SuccessIndexItem` to look if we don't know the sha? I was thinking that perhaps it should be declared as optional in that interface and then to not to include it if either `sha.txt` exists and contains the empty string or doesn't exist at all (as you prefer).

src/interpret.ts
```diff
@@ -371,11 +371,15 @@ function createResultIndex(
...
-    ) {
+    try {
       sha = fs.readFileSync(shaPath, "utf-8");
+    } catch (err: any) {
+      // check if this is a file not found error
+      if (err?.code === "ENOENT") {
+        console.log("File not found!");
+      } else {
+        throw err;
+      }
```
rneatherway, line 382: @charisk what do you think here? With the exists check first CodeQL warned of a race condition, so we changed it to catch the exception if the file doesn't exist instead. On the other hand this block feels very clunky. Perhaps we should just keep it simple and swallow any exception?

charisk: To me it looks like it doesn't matter what went wrong, but just that _something_ went wrong. So it'd make sense to try/catch any exception, maybe log it as a warning but continue normal operation. We'll then assume HEAD if the SHA isn't available in the extension code. 

If that turns out too problematic in practise we can change it.

shati-patel: Thanks! Something like this? https://github.com/dsp-testing/qc-run2/pull/639/commits/fd875ca0fb62d00fa9b9fb86fa2d9f1e91d886b8

rneatherway (approved):
:tada:

shati-patel:
Thanks for the reviews! Just tested with VS Code again, so I'll merge once everything passes 🧪

PR #642, opened 2022-03-02
[WIP] Include GitHub URL in SARIF locations

shati-patel:
Alternative to #639. Maybe we can directly embed the GitHub URL in the SARIF location.


shati-patel:
Best not to mess with SARIF. Going back to the other PR!

PR #648, opened 2022-03-04
Specify Accept: application/zip when downloading dbs

rneatherway:
This is required before we switch the default response to JSON.

Run: https://github.com/dsp-testing/qc-controller/runs/5425002903?check_suite_focus=true
I also checked the accept header was received server side.


robertbrignull (approved):
👍

PR #655, opened 2022-03-07
Add commit SHA + GitHub URL to SARIF output

shati-patel:
Add information to [`versionControlProvenance`](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Ref511829897) in the SARIF output.


src/codeql.ts
```diff
@@ -104,6 +104,8 @@ async function runQuery(
     outputSarif(
       codeql,
       bqrs,
+      nwo,
+      dbMetadata.creationMetadata?.sha || "HEAD",
```
rneatherway, line 108: Could we pass this in as a `string?` instead and then not write the `revisionId` field if it isn't available? I think that would be a little nicer than writing `HEAD`. The extension could then take the decision to use `HEAD` in constructing the links if it doesn't find `revisionId` in the downloaded SARIF file.

rneatherway (commented):
One small comment. It'd also be worth trying the produced SARIF file in https://sarifweb.azurewebsites.net/ as an extra safety net.

src/codeql.ts
shati-patel: Thanks! Attempted to do this in https://github.com/dsp-testing/qc-run2/pull/655/commits/4b53aeeaeca3dbdbf8acd4a0bb09f9432c46d4e5 🙇🏽‍♀️

rneatherway (approved):
:ship: :tada:

PR #661, opened 2022-03-08
Re-introduce code snippets in SARIF output

charisk:
These were removed in https://github.com/dsp-testing/qc-run2/pull/558 but it looks like it'd be useful to have both contents and snippets.


rneatherway (approved):

shati-patel (approved):
--sarif-add-ALL-the-things 👌🏽

PR #687, opened 2022-03-18
Avoid using codeql-action to initialise codeql

robertbrignull:
Swap out the call to `codeql-action/init` for just using the CodeQL bundle built in to the actions images. It's known that this won't work on self-hosted runners, but that's a concern for the future and we're happy to just deal with it later.

This should theoretically be all that's needed to use a controller repository that doesn't have GHAS enabled.


robertbrignull:
Tested just now and seems to work with a private controller repo
Screenshot 2022-03-21 at 10 22 26

I also tried it with more repos and a public controller repo, and those also worked just fine.

rneatherway (approved):

PR #695, opened 2022-03-24
Get a more accurate results count

shati-patel:
See internal issue for details.

If SARIF exists (i.e. for problem/path-problem queries), we get the count from the SARIF file. For non-problem queries, we still use the old method of counting in BQRS.

----- 
Tested in VS Code. Number of results now matches the SARIF output (and also matches the local query run) 🙌🏽 🎉 

<details><summary>remote</summary>
<p>

image

</p>
</details> 

<details><summary>local</summary>
<p>

image

</p>
</details>


src/codeql.ts
```diff
@@ -245,9 +247,22 @@ async function outputMd(
   );
   return [md];
 }
+/**
+ * Checks if the query kind is compatible with SARIF output.
+ */
+function queryCanHaveSarifOutput(compatibleQueryKinds: string[]): boolean {
+  if (
```
robertbrignull, line 254: I know I wrote it like this, but it would be slightly cleaner to remove the explicit `return true` and instead just do
```typescript
return compatibleQueryKinds.includes("Problem") || compatibleQueryKinds.includes("PathProblem");
```

robertbrignull (approved):
One minor suggestion but otherwise LGTM

shati-patel:
(Closed and reopened to kick off PR checks. Though I'll hold of on merging this until next week anyway.)

PR #699, opened 2022-03-28
Add permissions blocks to workflow jobs

robertbrignull:
We can limit the scopes that the actions token has, rather than giving it full write permissions to the `github/codeql-variant-analysis-action` repository. I think that the only job that needs any permissions at all is when we create the issue at the end. The proof will be in testing it though.


robertbrignull:
Integration tests seem to work ok: https://github.com/github/codeql-variant-analysis-action/actions/runs/2053604196

shati-patel (approved):
Test looks good, so 🚢

PR #703, opened 2022-03-29
Don't create an issue in the controller repo

robertbrignull:
This PR removes the behaviour of creating an issue in the controller repository containing all results. This was always a temporary implementation, and we can now remove it because the experience in vscode is complete enough that the issue is no longer needed.

This may conflict semantically with https://github.com/github/codeql-variant-analysis-action/pull/699 but I can deal with that depending which one merges first.


robertbrignull:
Integration tests seemed to have passed: https://github.com/github/codeql-variant-analysis-action/actions/runs/2058911312

It's slightly harder to tell now that it doesn't create an issue, but it has created the `result-index` upload so I think that means it all worked.

PR #711, opened 2022-03-30
Remove mention of MacOS from ci error

rneatherway:
I don't believe that this is correct since #440


robertbrignull (approved):
Makes sense to me

PR #713, opened 2022-04-04
Diagnostic query results are _not_ Problems

rneatherway:


rneatherway:
There are a couple of other places that check the kind or compatible kinds. One will be removed by #714, but we should probably unify any others.

charisk (approved):

robertbrignull (approved):
LGTM too

rneatherway:
The last remaining other place that tests against `Problem` kind is at https://github.com/github/codeql-variant-analysis-action/blob/c186865f80ba725cbcf33f3352bae09f31c57bad/src/codeql.ts#L209

That is safe because it's guarded by the function that I'm updating. However, it does seem like it could benefit from a small refactor. Perhaps `queryCanHaveSarifOutput` could return the inferred kind and that could then be passed in to `outputSarifAndCount` as well as guarded whether it is called.

PR #716, opened 2022-04-20
Revert check of diagnostic kind

rneatherway:
From #713. Unfortunately problem queries with result type
`Entity,String` are compatible with Diagnostic, so this prevents the
results being converted to SARIF.


shati-patel (approved):
Thanks!!! Tested with this branch and we have SARIF again 🏆

PR #717, opened 2022-05-04
Only download the query pack once

robertbrignull:
I noticed that we are downloading the query pack inside the loop over repositories. I don't think it's necessary to do it multiple times since the query pack is not modified by or between the queries.

I'm actually slightly confused why it wasn't previously causing an error, because the `download` function appears to throw an error if the destination file already exists 🤔


robertbrignull:
> I'm actually slightly confused why it wasn't previously causing an error

It wasn't failing because we create a temp workdir for each repository.

robertbrignull:
Here's a run over 1000 repos on `main` (https://github.com/dsp-testing/qc-controller/actions/runs/2269283600). You can see in the logs it prints "getting query pack" four times.

Here's a similar run on 1000 repos using this branch (https://github.com/dsp-testing/qc-controller/runs/6288556965?check_suite_focus=true) and you can see it only prints "getting query pack" once.

charisk (approved):
🚢

PR #718, opened 2022-05-05
Use download URL directly if one is provided

robertbrignull:
The idea here is that the instructions file could contain the download URLs directly rather than including a signed auth token that we then exchange for a download URL.

In this PR we add `downloadURL` to the `Repo` type and will use it if the value is provided. This should be backwards compatible, so we can merge this first, then make the change in the queries API to generate the download URLs (PR will come), and then remove the support here for the SATs and PATs.


robertbrignull:
We'll need to merge this before the backend change. If we're happy with the general idea of this proposed solution, then I suggest we merge this soon and then work on the other PR and discuss the details of it there.

robertbrignull:
Integration tests passed, though there isn't currently anything that touches this code path. We ideally need to add some more tests in this repo in general. Hopefully this PR is simple enough that we're happy just reviewing it and can be confident it won't break anything.

shati-patel (approved):
Changes here look sensible! 🤞🏽

robertbrignull:
I think let's merge this now well before the other PR and thus we'll be confident that this change itself hasn't broken anything, and if it does break anything we'll spot it and be able to revert this.

src/query.ts
```diff
@@ -15,6 +15,9 @@ import { download } from "./download";
 interface Repo {
   id: number;
   nwo: string;
+  downloadUrl?: string;
+
+  // token and pat are deprecated
   token?: string; // SignedAuthToken
   pat?: string;
```
rneatherway, line 22: `pat` is currently used (only) by the integration tests to download databases specified by the `instructions.json` contained in a release on this repository: https://github.com/github/codeql-variant-analysis-action/releases/tag/test

```json
{
    "repositories": [
        {
            "id": 376068344,
            "nwo": "dsp-testing/qc-demo-github-certstore"
        },
        {
            "id": 211169016,
            "nwo": "github/vscode-codeql"
        }

    ],
    "repoNwoChunks": [
        [
            "dsp-testing/qc-demo-github-certstore",
            "github/vscode-codeql"
        ]
    ]
}
```

One repository is public, one is private and will always require the bot PAT. Or we could definitely discuss other ways of integration testing!

robertbrignull: Good point we'll have to keep the path for calling the `GET .../databases/:language` endpoint to support the integration tests. That's a little sad that the tests will be using a slightly different code path that the production case, but it will be limited to a `if` block of only a few lines, so maybe it'll be acceptable, and it's still a HTTP request but just to a different place. 🤔 

For now I'm not worried, but we'll just have to keep the code for handling the PAT around when we clean this area up later.

PR #719, opened 2022-05-11
Remove references to SATs

robertbrignull:
Once the change in the API has been fully rolled out, these tokens will no longer be provided in the instructions file and we can remove support for them from the action.


robertbrignull:
Should be ready for review and merge now. The SATs should no longer be present in the instructions file anymore. This can be merged before or after the API change to fully remove mentions of SATs.

charisk (approved):

PR #720, opened 2022-05-13
Add a label to PRs once tests have passed once

robertbrignull:
This is an idea to reduce the chance of forgetting to run the integration tests. It's not perfect but hopefully it'll do enough to make us not forget.

The idea is that you're most likely to just forget to run the tests at all. Once you've done it once by updating the action branch and running the tests you're more likely to remember to run them again if you make any big changes to the PR. Therefore once the tests have run once with the correct action branch it adds a label to the PR. Then there's another CI check that checks for the presence of the label.

It's worth noting that this approach isn't perfect. It won't make you run the tests again if you make any changes. There are other approaches we could take to solve the problem of it being tedious to run the tests and easy to forget. This is just one approach and a fairly simple one. I think it's worth trying this since it's quite low effort, rather than putting lots of effort into doing something more full featured 🤷

### DIFF
--- a/.github/workflows/integration-test-call.yml
+++ b/.github/workflows/integration-test-call.yml
@@ -47,7 +47,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
 
       - name: Add success label to pull request
-        if: ${{ needs.query.result == 'success' && github.event.inputs.pr_number != "" }}
+        if: ${{ needs.query.result == 'success' && github.event.inputs.pr_number != '' }}
         run: |
           gh api -X POST "/repos/$GITHUB_REPOSITORY/issues/${{ github.event.inputs.pr_number }}/labels" \
             -d '{"labels":["integration tests succeeded"]}'
--- a/.github/workflows/integration-test-call.yml
+++ b/.github/workflows/integration-test-call.yml
@@ -53,8 +53,6 @@ jobs:
         if: ${{ needs.query.result == 'success' && github.event.inputs.pr_number != '' }}
         run: |
           gh pr edit "${{ github.event.inputs.pr_number }}" --add-label "integration tests succeeded"
-        env:
-          GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
 
       - name: Update status check to state 'failure'
         if: ${{ needs.query.result != 'success' }}
--- a/.github/workflows/integration-test-call.yml
+++ b/.github/workflows/integration-test-call.yml
@@ -35,6 +35,9 @@ jobs:
     needs:
       - query
     steps:
+      # Need to check out repo so that 'gh pr' will know the correct repo
+      - uses: actions/checkout@v2
+
       - name: Update status check to state 'success'
         if: ${{ needs.query.result == 'success' }}
         run: |
@@ -52,8 +55,6 @@ jobs:
           gh pr edit "${{ github.event.inputs.pr_number }}" --add-label "integration tests succeeded"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          OWNER: ${{ github.repository_owner }}
-          REPO: ${{ github.event.repository.name }}
 
       - name: Update status check to state 'failure'
         if: ${{ needs.query.result != 'success' }}
--- a/.github/workflows/integration-test-call.yml
+++ b/.github/workflows/integration-test-call.yml
@@ -25,7 +25,7 @@ jobs:
   query:
     needs:
       - init
-    uses: github/codeql-variant-analysis-action/.github/workflows/codeql-query.yml@main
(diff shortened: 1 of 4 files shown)


robertbrignull:
Ok, it needs to detect what the correct branch is for the PR, rather than trying to use the merge ref.

robertbrignull:
The `integration-tests-have-been-run` check has failed as intended. I'll now try running the integration tests.

robertbrignull:
Ok, a successful run at last and it's added the label.

robertbrignull:
Yay! I think it all works now. Only took about 12 attempts 😢

.github/workflows/integration-test-call.yml
```diff
@@ -34,6 +35,9 @@ jobs:
     needs:
       - query
     steps:
+      # Need to check out repo so that 'gh pr' will know the correct repo
+      - uses: actions/checkout@v2
```
adityasharad, line 39: Minor: we can start using v3

.github/workflows/integration-test-trigger.yml
```diff
@@ -2,6 +2,8 @@ name: "Integration test (trigger only)"
 
 on:
   push:
+    branches: [main]
+  pull_request:
```
adityasharad, line 6: Do you want to filter this to PRs against `main`?

.github/workflows/integration-test-trigger.yml
```diff
@@ -21,11 +23,17 @@ jobs:
       # The owner of the provided token must have write access to trigger this workflow run.
       - name: Trigger the real workflow
         run: |
+          if [ ${{ github.event_name }} == "pull_request" ]; then
```
adityasharad, line 26: Most of these contexts have an equivalent default environment variable, and I generally find it safer to use the variable when interpolated into Bash.
```suggestion
          if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
```

.github/workflows/integration-test-trigger.yml
```diff
@@ -21,11 +23,17 @@ jobs:
...
+            REF="${{ github.event.pull_request.head.ref }}"
```
adityasharad, line 27:
```suggestion
            REF="$GITHUB_HEAD_REF"
```

adityasharad (approved):
Neat. Some minor suggestions around variable use.
We could make this more involved and have the integration tests create their own check runs on the PR, but that machinery is somewhat complex to get right. Go with this and see how well it works!

.github/workflows/integration-test-trigger.yml
robertbrignull: I think we want to run against all PRs. That gives more immediate feedback even if you're merging into another feature branch. Shouldn't make too much difference in practice. The `integration-tests-have-been-run` is also not a required check currently.

.github/workflows/integration-test-call.yml
```diff
@@ -45,6 +49,13 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
 
+      - name: Add success label to pull request
+        if: ${{ needs.query.result == 'success' && github.event.inputs.pr_number != '' }}
+        run: |
+          gh pr edit "${{ github.event.inputs.pr_number }}" --add-label "integration tests succeeded"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
aeisenberg, line 57: Minor: this is what we've tended to do in our other workflows:

```suggestion
        run: |
          gh pr edit "$PR_NUM" --add-label "integration tests succeeded"
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          PR_NUM: ${{ github.event.inputs.pr_number }}
```

But, what you have is fine.

.github/workflows/pr-checks.yml
```diff
@@ -16,6 +16,17 @@ jobs:
         run: |
           ! grep 'uses: github/codeql-variant-analysis-action/' .github/workflows/codeql-query.yml | grep -q -v '@main$'
 
+  integration-tests-have-been-run:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check label is present on PR
+        if: ${{ "$GITHUB_EVENT_NAME" == 'pull_request' }}
```
aeisenberg, line 24: Minor: If this works, then I have no problem with it. I didn't know that we could use environment variables in `if` expressions.

```suggestion
        if: ${{ github.event_name == 'pull_request' }}
```

aeisenberg (approved):
These are both minor comments and this can go in as-is. In our other workflows, we tend to avoid workflow expressions in scripts. This is more for consistency and style, not for correctness.

.github/workflows/pr-checks.yml
robertbrignull: Good point. I'd forgotten this was a special actions `if` expression.

PR #722, opened 2022-05-17
Don't group results by location

shati-patel:
When we have multiple results for a single location, the alert message gets repeated multiple times, which looks odd:
image

Let's see if passing in the [`--no-group-results`](https://codeql.github.com/docs/codeql-cli/manual/bqrs-interpret/#cmdoption-codeql-bqrs-interpret-no-group-results) flag when interpreting BQRS fixes the issue!

~Note: I hope this will Just Work™, but we may need to make changes to how we process the SARIF in VS Code. I'll leave this in draft mode while I do some testing.~ 

🌟 Tested in VS Code and all looks fine. See linked internal issue for example results 🌟


robertbrignull (approved):
So now it appears as three separate results with the same message, rather than one alert with a mangled message? Sounds good to me

aeisenberg (approved):
I think this is reasonable.

PR #723, opened 2022-05-17
Adjustments to integration tests label check

robertbrignull:
Hopefully fixed the problems from https://github.com/github/codeql-variant-analysis-action/pull/720

The `integration-tests-have-been-run` check now runs when a label is added to the PR so it'll update its state.

The "Add success label to pull request" step now only adds the label if the workflow references the current branch.

### DIFF
--- a/.github/workflows/integration-test-call.yml
+++ b/.github/workflows/integration-test-call.yml
@@ -60,7 +60,6 @@ jobs:
             gh pr edit "$PR_NUM" --add-label "integration tests succeeded"
           else
             echo "Detected $NUM_USAGES references to $REF branch"
-            exit 1
           fi
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/integration-test-call.yml
+++ b/.github/workflows/integration-test-call.yml
@@ -52,7 +52,16 @@ jobs:
       - name: Add success label to pull request
         if: ${{ needs.query.result == 'success' && github.event.inputs.pr_number != '' }}
         run: |
-          gh pr edit "$PR_NUM" --add-label "integration tests succeeded"
+          # Get the branch without refs/heads/
+          REF=$(git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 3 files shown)


robertbrignull:
So https://github.com/github/codeql-variant-analysis-action/runs/6471373184?check_suite_focus=true said "detected 0 references" so that's good, and it didn't add the label. The label check is also failing because there is no label.

All good so far.

robertbrignull:
Adjusted check of ref to remove the `refs/heads/` and force pushed back to start to test the full flow.

robertbrignull:
Ok now posts "Detected 0 references to robertbrignull/tests_fix branch" as expected

.github/workflows/integration-test-call.yml
```diff
@@ -52,7 +52,15 @@ jobs:
       - name: Add success label to pull request
         if: ${{ needs.query.result == 'success' && github.event.inputs.pr_number != '' }}
         run: |
-          gh pr edit "$PR_NUM" --add-label "integration tests succeeded"
+          # Get the branch without refs/heads/
+          REF=$(git rev-parse --abbrev-ref HEAD)
+          # Only add the label if the workflow references the correct branch
+          NUM_USAGES=$(grep 'uses: github/codeql-variant-analysis-action/' .github/workflows/codeql-query.yml | grep "@${REF}$" | wc -l)
```
aeisenberg, line 58: minor: I think this is fine, but I prefer quoting all variable assignments in bash:

```suggestion
          REF="$(git rev-parse --abbrev-ref HEAD)"
          # Only add the label if the workflow references the correct branch
          NUM_USAGES="$(grep 'uses: github/codeql-variant-analysis-action/' .github/workflows/codeql-query.yml | grep "@${REF}$" | wc -l)"
```

I don't think it matters in this case, but I do it out of habit since I've been bitten by this too many times. This would use nested quotes but that is fine.

aeisenberg (approved):
My comment is minor, and just a stylistic suggestion. Looks good as-is.

PR #724, opened 2022-05-23
Refactor: Return nwo/count/sha in a single metadata file

shati-patel:
We currently pass metadata from the `run-query` step to the `combine-results` step by uploading an artifact containing text files with each bit of info (e.g. `nwo.txt`, `sha.txt`, `resultcount.txt`). The `combine-results` step then uses that metadata to create a result index file. 

To make this neater, this PR combines them into a single `metadata.json` file. There should be no visible behaviour change.

See internal issue for more context.

-----

Question: We still have a separate `error.txt` file for failed runs - should we also replace that with an entry in the metadata file?


src/codeql.ts
```diff
@@ -360,3 +353,27 @@ async function getRemoteQueryPackDefaultQuery(
 
   return queries[0];
 }
+
+export interface QueryRunMetadata {
```
charisk, line 357: What do you think about moving all query run metadata code into its own module? It's currently spread between `codeql.ts` and `interpret.ts` and I'm thinking it doesn't have much to do with CodeQL.

We could have a `query-run-metadata.ts` that includes the `QueryRunMetadata` interface as well as `writeQueryRunMetadataToFile()` and `readQueryRunMetadataFromFile()` (currently called `readMetadata()`).

Note that if you make this change I'd suggest updating the `readQueryRunMetadataFromFile()` to receive a file path (the `response.downloadPath`) rather than the whole `DownloadResponse` object. That way the query run metadata module is agnostic of any response stuff.

charisk (commented):
Looking good! Some small comments from me - I'll let someone else review since we paired on this a fair bit and I'm not massively familiar with the repo.

robertbrignull (commented):
Looks good to me too, and I agree with everything @charisk said. 

> Question: We still have a separate `error.txt` file for failed runs - should we also replace that with an entry in the metadata file?

Yeah we definitely could. Currently we detect success/failure by whether `error` is in the artifact name, but that easily be checking for a field in the `metadata.json` instead. I wouldn't worry much about the inefficiency of having to read the metadata file as we'd have to read it later anyway. I think having the `metadata.json` file and then also optionally the results files sounds good, and therefore also keep the artifact name always just the repo id. Doing this in a separate PR might be a good idea though since it actually looks like it'll change quite a bit.

charisk (approved):

PR #726, opened 2022-05-30
Update check-node-modules.sh to mention running on linux

robertbrignull:
What do people think of this? Is this enough of a warning to help us remember how to do things?

I'm also not actually sure if linux is strictly needed as windows might work too. All I know currently is that mac doesn't work.

### DIFF
--- a/.github/workflows/script/check-node-modules.sh
+++ b/.github/workflows/script/check-node-modules.sh
@@ -15,6 +15,7 @@ npm run removeNPMAbsolutePaths
 if [ ! -z "$(git status --porcelain)" ]; then
     # If we get a fail here then the PR needs attention
     >&2 echo "Failed: node_modules are not up to date. Run 'npm ci && npm run removeNPMAbsolutePaths'."
+    >&2 echo "For best results use linux when running the above commands. Use a codespace or docker container if necessary."
     git status
     exit 1
 fi


shati-patel (approved):
LGTM

aeisenberg:
Mac works fine when running this command on the codeql-action. Is there something different about this action?

robertbrignull:
I'm not sure. Definitely a lot of inspiration for the structure of this repo was taken from the codeql-action. Could the codeql-action have advanced since then and fixed this?

Currently if I run `npm ci && npm run removeNPMAbsolutePaths` on a mac it tries to install `fsevents`.
```
> git status
On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	node_modules/fsevents/

nothing added to commit but untracked files present (use "git add" to track)
```

aeisenberg:
I have no power at my house now and I'm running tethered to my phone, so I can't check this out myself on the codeql-action.

I _think_ I know what's happening. Locally, I never run `npm ci`. If I need to add or change dependencies, I just run `npm install -S foo && npm run removeNPMAbsolutePaths`. This seems to work for me. So, I never hit the problem above since I use a different workflow.

PR #727, opened 2022-05-30
Clean-up: Delete unused inputs from `combine-results`

shati-patel:
Since https://github.com/github/codeql-variant-analysis-action/pull/703, `combine-results` no longer uploads an issue. That PR removed any dependence on the action inputs, so I think we can remove them 💥


robertbrignull (approved):
Yep, the only calls to `getInput` that I can see are from `query.ts`

PR #729, opened 2022-06-21
Clear up files between queries to help avoid running out of disk space

robertbrignull:
We create a new working directory for each repository that we analyse, but we do not clear up these directories between queries. This can lead to running out of disk space if four queries use a lot of space. We can help avoid this and thus we'll only fail if a query uses all the space on its own.


robertbrignull:
Here's my suggestion for deleting the work directory between queries.

Unfortunately I haven't been able to reproduce the out-of-disk error. I tried creating a custom `instructions.json` file that would run the `mp911de/logstash-gelf, RestExpress/RestExpress, poetix/protonpack, Blazebit/blaze-persistence` repositories in one job, and these are the repos that caused the error in https://github.com/dsp-testing/qc-controller/actions/runs/2344503810. Unfortunately after trying it a few times it always completed with loads of free space. I added in a call to `df` and it showed 31GB of free space. This leads me to think there was something else going on which caused the disk space issues earlier, but I'm not sure what.

However, I think doing this change is likely still beneficial. Just from a logical perspective it makes sense to delete these files once we're done with them and prevent them from interfering with future queries.

I did some integration test runs (e.g. https://github.com/github/codeql-variant-analysis-action/actions/runs/2534864625) and you can see that the artifact uploads are still present, so deleting the files after uploading them is no problem.

charisk (approved):
🚢

PR #730, opened 2022-06-22
Configure available RAM for running queries

shati-patel:
Explicitly set the `--ram` flag for [`codeql database run-queries`](https://codeql.github.com/docs/codeql-cli/manual/database-run-queries/#options-to-control-ram-usage), so that codeql uses all available memory. 

This should help avoid the OOM errors that some users have seen.


robertbrignull (approved):
Looks correct to me 👍 

I'm happy with the logic because it's the exact as we do in the codeql-action, and the code itself looks good.

shati-patel:
> Looks correct to me 👍
> 
> I'm happy with the logic because it's the exact as we do in the codeql-action, and the code itself looks good.

Thanks! I'm wondering if it's worth adding tests or something? 🤔

robertbrignull:
> I'm wondering if it's worth adding tests or something? 🤔

Hmm, it may be difficult to write some because the output value depends on the system where the test is being run, and stubbing those bits to make the test consistent could be quite hard and annoying. I think you could get away without anything, but if anyone else has suggestions please do say.

shati-patel:
> > I'm wondering if it's worth adding tests or something? 🤔
> 
> Hmm, it may be difficult to write some because the output value depends on the system where the test is being run, and stubbing those bits to make the test consistent could be quite hard and annoying. I think you could get away without anything, but if anyone else has suggestions please do say.

Okay, I've tested a few MRVA runs from VS Code and everything seems fine, so I'm happy enough to merge this as-is 🙇🏽‍♀️ 

(Will just change the testing branch back to `main`)

PR #733, opened 2022-08-15
Make return type of `runQuery` more flexible

shati-patel:
For work around "live results", we'll need to upload results in a different format. This is some preparatory work to give us more flexibility in accessing results from running a query.

See internal issue for more details!


src/codeql.ts
```diff
@@ -21,29 +21,36 @@ export {
 // This name must match that used by the vscode extension when creating the pack.
 const REMOTE_QUERY_PACK_NAME = "codeql-remote/query";
 
+interface RunQueryResult {
```
charisk, line 24: We should probably export this interface considering it's part of the public interface of this module.

charisk (approved):
LGTM but might want to get eyes from someone else considering we paired on most of this.

robertbrignull (approved):
Looks good to me too. I really like making the return type more explicit rather than just a list of strings.

PR #734, opened 2022-08-15
Split out function for getting BQRS file

shati-patel:
The `runQuery` function had become quite long and complicated. I've split out one bit that I found particularly confusing into a separate function, which should help declutter the main `runQuery` logic.

Alternative suggestions always welcome! 🌻


src/codeql.ts
```diff
@@ -370,3 +352,36 @@ async function getRemoteQueryPackDefaultQuery(
 
   return queries[0];
 }
+
+/**
+ * Checks that `codeql database run-queries` produced exactly one BQRS result file.
```
charisk, line 357: How about this for a description: `Finds the BQRS file fora database and ensures that exactly one is produced` or something like that?

charisk (approved):
LGTM, just a minot suggestion.

Thanks for making this change!

src/codeql.ts
```diff
@@ -75,26 +74,9 @@ async function runQuery(
...
-
-  const entry = entries[0];
-  if (!entry.isFile() || !entry.name.endsWith(".bqrs")) {
-    throw new Error(`Unexpected file in ${cur}: ${entry.name}`);
-  }
-
-  fs.renameSync(path.join(cur, entry.name), bqrsFilePath);
+  const bqrsFilePath = path.join("results", "results.bqrs");
+  const tempBqrsFilePath = getBqrsFile(databaseName);
+  fs.renameSync(tempBqrsFilePath, bqrsFilePath);
```
robertbrignull, line 79: I dithered on posting a comment here about moving the rename inside the `getBqrsFile` method, but eventually decided it doesn't make much difference. So I wanted to note this as a point but I'm not requesting any changes.

The crucial point for me is that the `getBqrsFile` method isn't generating the BQRS file but it's just finding an existing file that CodeQL created earlier. So I think things make sense as they are and keeping the `rename` outside the method is good.

robertbrignull (approved):
Changes LGTM ✨

src/codeql.ts
shati-patel: Yeah, FWIW I went through a similar thought process myself 😂 I think either way would be okay 🤷🏽

src/codeql.ts
shati-patel: Thanks, done! 👍🏽

PR #736, opened 2022-08-25
Create temporary directory for each repo when uploading errors

shati-patel:
Avoid `EEXIST` error when uploading error artifacts for multiple repos in the same job. 

In the for loop where we actually run the query, we make sure to `cd` into a fresh directory for each repository, so that the created artifacts don't conflict. I've copied that same approach to the for loop where we handle earlier errors. 

See internal issue for more details! 🦒


shati-patel:
(Will fix `PR checks / workflow-references-main` once this PR has been reviewed!)

robertbrignull (approved):
Looks good to me, though I helped write some of it.

I kinda wish we could remove the whole process of changing working directory, but for this PR I wouldn't change any of that. This looks good as it is. We fix the bug by using `uploadError` in the way it expects to be used.

PR #737, opened 2022-09-05
Implement GitHub API client for "live results" endpoints

shati-patel:
Currently unused, but we'll hook these up in a later stage. 

I'm not sure if/how to test these yet. Perhaps we want to add some mock data or responses once we actually call these functions? 🤔


src/api-client.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  const url = `/codeql/variant-analyses/${variantAnalysisId}/repositories/${repoId}`;
+  const response = await http.patch(url, JSON.stringify(data));
+  if (response.message.statusCode !== 204) {
+    console.log(
+      `Request to ${url} returned status code ${response.message.statusCode}:
+      ${await response.readBody()}`
+    );
+    throw new Error(
+      `Error while setting variant analysis as "${status}". Status code: ${response.message.statusCode}`
```
robertbrignull, line 82:
```suggestion
      `Error while setting variant analysis as "${data.status}". Status code: ${response.message.statusCode}`
```

Ooops, missed this one. Turns out we didn't notice and it continued to compile because `status` is a [built-in thing for javascript](https://www.w3schools.com/jsref/prop_win_status.asp), but it's not what we want. If we were running in a browser, we could set some text on the status bar 🤷

robertbrignull (approved):
One comment on a bit we missed but LGTM otherwise! 🚢 🇮🇹

src/api-client.ts
shati-patel: Woops 🙈 Good spot, thanks!!!

src/api-client.ts
```diff
@@ -0,0 +1,114 @@
...
+
+async function updateVariantAnalysisStatus(
+  variantAnalysisId: number,
+  repoId: number,
+  data: updateVariantAnalysisProperties
+): Promise<void> {
+  const http = getApiClient();
+
+  const url = `/codeql/variant-analyses/${variantAnalysisId}/repositories/${repoId}`;
+  const response = await http.patch(url, JSON.stringify(data));
```
koesie10, line 75: Would it make sense to use `patchJson` instead? That should automatically set the `Content-Type` header (which might be necessary for the GitHub API, I'm not sure), parse the response as JSON and reject the promise when the request returns a non-200 status code (which will throw an error since you're awaiting the call).

src/api-client.ts
```diff
@@ -0,0 +1,114 @@
...
+  resultCount: number;
+  databaseSHA: string;
+}
+
+interface FailureProperties {
+  status: "failed";
+  failureMessage: string;
+}
+
+type updateVariantAnalysisProperties =
```
charisk, line 27: Lint police :D Types should start with capital letter.

src/api-client.ts
```diff
@@ -0,0 +1,114 @@
+import * as httpm from "@actions/http-client";
+
+const userAgent = "GitHub multi-repository variant analysis";
+
+export function getApiClient() {
+  return new httpm.HttpClient(userAgent, [], {
+    allowRetries: true,
```
charisk, line 7: Just curious, what's the default behaviour for this? Is it exponential retries?

src/api-client.ts
```diff
@@ -0,0 +1,114 @@
+import * as httpm from "@actions/http-client";
+
+const userAgent = "GitHub multi-repository variant analysis";
```
charisk, line 3: Should this include information about the fact that we're calling from the Action workflow? E.g. `GitHub multi-repository variant analysis action`

src/api-client.ts
```diff
@@ -0,0 +1,114 @@
+import * as httpm from "@actions/http-client";
+
```
charisk, line 2: What do you think about calling this file `gh-api-client` to make it more explicit about what it actually is?

src/api-client.ts
```diff
@@ -0,0 +1,114 @@
...
+    throw new Error(
+      `Error while setting variant analysis as "${data.status}". Status code: ${response.message.statusCode}`
+    );
+  }
+}
+
+export async function getPolicyForRepoArtifact(
+  variantAnalysisId: number,
+  repoId: number,
+  artifact_size: number
```
charisk, line 90: Should be `artifactSize` (this is not Ruby m'am 😹)

charisk (commented):
Looking good, some minor comments.

src/api-client.ts
robertbrignull: Not in this PR as it might change a lot of things, but shall we enable https://typescript-eslint.io/rules/naming-convention and then it'll be linted for us?

robertbrignull: For the record, if you turn it on it finds this:
```
/Users/robertbrignull/github/codeql-variant-analysis-action/src/api-client.ts
  27:6  error  Type Alias name `updateVariantAnalysisProperties` must match one of the following formats: PascalCase  @typescript-eslint/naming-convention
  90:3  error  Parameter name `artifact_size` must match one of the following formats: camelCase                      @typescript-eslint/naming-convention
  94:5  error  Object Literal Property name `content_type` must match one of the following formats: camelCase         @typescript-eslint/naming-convention

/Users/robertbrignull/github/codeql-variant-analysis-action/src/interpret.ts
  14:3   error  Type Property name `results_count` must match one of the following formats: camelCase                     @typescript-eslint/naming-convention
  15:3   error  Type Property name `bqrs_file_size` must match one of the following formats: camelCase                    @typescript-eslint/naming-convention
  16:3   error  Type Property name `sarif_file_size` must match one of the following formats: camelCase                   @typescript-eslint/naming-convention
  17:3   error  Type Property name `source_location_prefix` must match one of the following formats: camelCase            @typescript-eslint/naming-convention
  50:11  error  Variable name `bqrs_file_size` must match one of the following formats: camelCase, UPPER_CASE             @typescript-eslint/naming-convention
  53:9   error  Variable name `sarif_file_size` must match one of the following formats: camelCase, UPPER_CASE            @typescript-eslint/naming-convention
  63:7   error  Object Literal Property name `results_count` must match one of the following formats: camelCase           @typescript-eslint/naming-convention
  64:7   error  Object Literal Property name `bqrs_file_size` must match one of the following formats: camelCase          @typescript-eslint/naming-convention
  65:7   error  Object Literal Property name `sarif_file_size` must match one of the following formats: camelCase         @typescript-eslint/naming-convention
  66:7   error  Object Literal Property name `source_location_prefix` must match one of the following formats: camelCase  @typescript-eslint/naming-convention

✖ 13 problems (13 errors, 0 warnings)
```

So most (but not all) of those are property names that will be serialized, so we can't change them super easily and might need to exclude all those from the check. But overall would still be useful to have this checked automatically.

src/api-client.ts
robertbrignull: I didn't immediately find docs on this (maybe there are some) but it does look like exponential backoff: https://github.com/actions/toolkit/blob/main/packages/http-client/src/index.ts#L432

src/api-client.ts
robertbrignull: Maybe, though the actual implementation seems to be a bit unusual so we might still need some custom error handling. Overall I think doing it manually might be just as easy, both in terms of amount of code and readability.

See [`patchJson`](https://github.com/actions/toolkit/blob/bc4be505973a6a7344bfd71e1b32f77e1755310c/packages/http-client/src/index.ts#L299-L321) and [`_processResponse`](https://github.com/actions/toolkit/blob/bc4be505973a6a7344bfd71e1b32f77e1755310c/packages/http-client/src/index.ts#L696-L769) for the code that implements this.

The behaviour that I see as odd is that in the case of a [404 response](https://github.com/actions/toolkit/blob/bc4be505973a6a7344bfd71e1b32f77e1755310c/packages/http-client/src/index.ts#L710) it resolves the promise as successful, but with a null `respnose.result` field. This is a bit unhelpful for us because we don't expect to hit 404s and if we do then we want it to be treated as an error.

Potentially we could rewrite as something like this:
```typescript
const response = await http.patchJson(url, data);
if (response.statusCode !== 204) {
  throw new Error(
    `Error while setting variant analysis as "${data.status}". Status code: ${response.statusCode}`
  );
}
```

What do other people think?

charisk: That's a bit strange. It sounds like it's not worth it to me, but no strong opinions.

src/api-client.ts
charisk: Yes please!

src/api-client.ts
shati-patel: Thanks for the comments! Since there's no super strong feelings, I've left this as-is for now. Perhaps something we could do in a follow-up PR if that makes sense 😃

robertbrignull (approved):
👍

src/gh-api-client.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const userAgent = "GitHub multi-repository variant analysis action";
+
+export function getApiClient() {
+  return new httpm.HttpClient(userAgent, [], {
+    allowRetries: true,
+  });
+}
+
+interface InProgressProperties {
```
charisk, line 11: These names have room for improvement. We're in the context of a gh-api-client so "InProgressProperties" sounds a bit too generic to me.

How about `InProgressAnalysis`, `SuccessfulAnalysis` etc.?

src/gh-api-client.ts
```diff
@@ -0,0 +1,114 @@
...
+  status: "in_progress";
+}
+
+interface SuccessProperties {
+  status: "succeeded";
+  sourceLocationPrefix: string;
```
charisk, line 17: Should these properties be `source_location_prefix` etc.? This is an interface representing the API model, which uses snake care rather than camel case.

robertbrignull: Good point, we didn't check these against the API schema closely enough 😢 

I've checked now and they should be:
- `source_location_prefix`
- `result_count`
- `database_commit_sha`

src/gh-api-client.ts
```diff
@@ -0,0 +1,114 @@
...
+  resultCount: number;
+  databaseSHA: string;
+}
+
+interface FailureProperties {
+  status: "failed";
+  failureMessage: string;
```
robertbrignull, line 24: This should be `failure_message` to match the API schema

charisk (approved):

PR #738, opened 2022-09-07
Implement Azure client for uploading result artifacts

shati-patel:
Adds a function that uploads result artifacts (either SARIF or BQRS, zipped) to Azure. This is currently unused, but we'll hook this up to the "real" code in a later step!

I've mostly followed the example from `@koesie10`'s ✨ investigation work (see internal issue), which included a useful new dependency (`form-data`).


src/azure-client.ts
```diff
@@ -0,0 +1,48 @@
+import * as fs from "fs";
+
+import FormData from "form-data";
+
+import { getApiClient } from "./gh-api-client";
+
+export interface Policy {
+  upload_url: string;
+  header: Map<string, string>;
+  form: Map<string, string>;
```
robertbrignull, line 10: We might have gone a bit overboard by declaring these as a `Map`. I wasn't sure if `Map` was just a synonym for `Object` but it is not. They're different and not really interchangeable. See https://medium.com/dailyjs/7-differences-between-objects-and-maps-in-javascript-bc901dfa9350 for some more info.

The particularly problematic bits for us are that:
- When you deserialise some JSON it'll always become an object. Ideally we'd like to able to just `JSON.parse` the result of the API request to get the policy.
- Later on when we do `...policy.header`, that doesn't work with a `Map`. It's not a compile error but it just does nothing.

Luckily we can make these fields objects but still specify types for them. I'm my opinion the syntax is a bit arcane and less readable, but it works. I suggest we replace these two lines with:
```suggestion
  header: [name: string]: string;
  form: [name: string]: string;
```
So this is still saying that our object has keys of type `string` and values of type `string`.

src/azure-client.ts
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+  const response = await httpClient.sendStream(
+    "POST",
+    policy.upload_url,
+    data,
+    additionalHeaders
+  );
+
+  if (!response.message.statusCode || response.message.statusCode > 299) {
```
robertbrignull, line 39: Just a note to other readers that this response is coming from azure / memory-alpha, so we aren't as directly in control of it. I don't currently know exactly what status code we get here, but I don't think it matters too much to specify exactly what we expect here. Anything in the 2xx range should indicate success. So I support just leaving this code as it is.

src/azure-client.ts
```diff
@@ -0,0 +1,48 @@
+import * as fs from "fs";
+
+import FormData from "form-data";
+
+import { getApiClient } from "./gh-api-client";
+
+export interface Policy {
+  upload_url: string;
```
robertbrignull, line 8: Note that this is deserialized from an earlier HTTP response so we don't control that this is snake case.

src/azure-client.ts
```diff
@@ -0,0 +1,48 @@
...
+  header: Map<string, string>;
+  form: Map<string, string>;
+}
+
+export async function uploadArtifact(policy: Policy, artifactZipPath: string) {
+  const data = new FormData();
+  for (const [key, value] of policy.form.entries()) {
```
robertbrignull, line 15: If you change the type of `policy.form` then this will need to change as follows:
```suggestion
  for (const [key, value] of Object.entries(policy.form)) {
```

src/azure-client.ts
koesie10: Instead of `[name: string]: string`, you can also use `Record<string, string>` which is equivalent but might be a bit easier to read.

shati-patel: Awesome, thanks both 😍 🙌🏽

shati-patel:
Thanks for the reviews! ✨

robertbrignull (approved):
I'm happy with this as it is now 👍

PR #739, opened 2022-09-07
Add ESLint rule to enforce naming conventions 

shati-patel:
As discussed in https://github.com/github/codeql-variant-analysis-action/pull/737#discussion_r963357745, it would be helpful to have a lint rule that picks up on naming conventions.

I've used the default settings with the exception of allowing `snake_case` for property names (since they need to match the required properties in the API). I'm open to suggestions on making this more/less specific! 🔍 

PS: Prettier really wanted to autoformat the `.eslintrc.json` file. I've done the formatting in https://github.com/github/codeql-variant-analysis-action/commit/f82f908b991aaa567da9647105ee030d50e9097f, and the ✨ actual change ✨ in https://github.com/github/codeql-variant-analysis-action/commit/f82f908b991aaa567da9647105ee030d50e9097f


robertbrignull:
Looks like a reasonable option though I'd like to explore the options on property names a bit more.

Do we have other property names that aren't used for talking to APIs or other systems that require snake case? It would nice if those were still covered by the naming convention rule so we don't start accidentally slipping up.

One option you see frequently is to enforce the rule everywhere but then use comments to individually allow specific cases. In this case we'd add `// eslint-disable-line @typescript-eslint/naming-convention` wherever we have a property name that breaks the rule. Would that be better?

charisk:
Sorry if I'm saying something obvious but there are other ways to temporarily disable an ESLint rule. `// eslint-disable-line` disables the next line but could also disable for a file, disable for a block, etc. See https://learn.coderslang.com/0023-eslint-disable-for-specific-lines-files-and-folders/

shati-patel:
Will come back to this shortly! As discussed offline, I'll move all the `snake_case` names into `gh-api-client.ts` and then disable the naming rule in that file only.

shati-patel:
> Will come back to this shortly! As discussed offline, I'll move all the `snake_case` names into `gh-api-client.ts` and then disable the naming rule in that file only.

Done! Two additional comments: 

- I've moved the `Policy` interface into `gh-api-client.ts`, which will cause a minor conflict with #741 
- There were some additional `snake_case`d properties in `interpret.ts`. These also come from API/extension, so I've disabled the naming rule for those specific blocks (see https://github.com/github/codeql-variant-analysis-action/pull/739/commits/892341fa71a3fdaf6040d741ae9489a7eda91409).

charisk (approved):

robertbrignull (approved):

PR #741, opened 2022-09-13
Update GitHub API client to use octokit

shati-patel:
This PR replaces the http client in `gh-api-client.ts` with octokit (using https://github.com/octokit/action.js), which allows us to authenticate more easily using built-in actions tokens and permissions. 

This code is not yet used anywhere, so if this looks safe, perhaps we could merge this and then test the changes (using a rebased version of https://github.com/github/codeql-variant-analysis-action/pull/740) 🤔 

See individual commits for more info! (most of the "Files changed" are just dependencies) 

**Note**: 
This change means we now have two clients: octokit + the old http client (used in `azure-client.ts` and `download.ts`). We'll replace everything with octokit eventually, but I'd rather do that in a separate PR 🙇🏽‍♀️


src/gh-api-client.ts
```diff
@@ -30,85 +35,104 @@ type UpdateVariantAnalysis =
...
   data: UpdateVariantAnalysis
 ): Promise<void> {
-  const http = getApiClient();
+  const octokit = getOctokit();
 
-  const url = `/codeql/variant-analyses/${variantAnalysisId}/repositories/${repoId}`;
-  const response = await http.patch(url, JSON.stringify(data));
-  if (response.message.statusCode !== 204) {
+  const url = `PATCH /repositories/${controllerRepoId}/code-scanning/codeql/variant-analyses/${variantAnalysisId}/repositories/${repoId}`;
+  const response = await octokit.request(url, { data });
```
charisk, line 99: Octokit throws exceptions for failure requests e.g. 404 so we'll need to have try/catch around `octokit.request` to capture those. Unfortunately simply checking the success code is not enough.

robertbrignull: Alternatively, do we need to capture the exceptions or just let them bubble up? If the exception is descriptive enough already to be useful in debugging then it's fine and we could remove all the status-checking code entirely.

shati-patel:
Just pushed a commit to resolve git conflicts - I'll address comments tomorrow 😎

src/gh-api-client.ts
shati-patel: I've removed the status-checking entirely, so hopefully octokit's errors will be descriptive enough.

(We can't really test until I've rebased https://github.com/github/codeql-variant-analysis-action/pull/740 onto this though 🙃 )

robertbrignull (approved):
I'm happy with this approach. @charisk, what do you think?

charisk (approved):
LGTM!

PR #742, opened 2022-09-14
Some more ESLint tidying

shati-patel:
Fixes an ESLint error on `main`, and tidies up the ESLint config file 🧹


robertbrignull (approved):
LGTM

Thanks for reminding me of those disabled rules and removing a couple of them. In the future we should look at removing a few more of those.

PR #744, opened 2022-09-19
Add post action for setting failure

koesie10:
This will set all repo tasks in a job as failed when the run failed or is cancelled. I couldn't find a way to retrieve the error message or the outcome in the post step, so it uses a generic message.


src/query.ts
```diff
@@ -140,6 +133,10 @@ async function run(): Promise<void> {
     // We can now delete the work dir. All required files have already been uploaded.
     chdir(curDir);
     fs.rmdirSync(workDir, { recursive: true });
+
+    // Save that we have already completed this repo so we don't set the state
+    // to failure in the post-action when a later repo fails.
+    saveState(`repo_${repo.id}_completed`, "true");
```
robertbrignull, line 139: I think we need to use `saveState` around https://github.com/github/codeql-variant-analysis-action/blob/56439308519c257fcebcafcb475b1dd34d1d7215/src/query.ts#L75 too. That code is handling if we hit an error downloading the query pack, so we mark the job as failed and update the state of each repo task. So if we do that there we don't want to do it again in the post action.

robertbrignull (approved):
Generally LGTM. I think there's one place we need to also consider saving state so we don't overwrite previous errors, but generally the approach looks great.

I like that we're sharing the code for reading the inputs, and using the existing `setVariantAnalysisFailed` method.

I see the concern about not knowing within `post-query.ts` whether the workflow was cancelled or failed.  Overall I think this is fine, at least for now. It's certainly something we can look into in the future, but let's wait to see if this causes problems for any users first.

src/query.ts
koesie10: Thanks! I've now added it and also moved this `saveState` call to be above the `chdir` call since we don't need to set the state to failed if that step fails (which is very unlikely anyway).

robertbrignull (approved):
👍

PR #745, opened 2022-09-20
Only upload actions artifacts when not in live results mode

robertbrignull:
When in live-results mode, we don't want to upload the old actions artifacts because they are unnecessary and also a privacy concern.

Shouldn't affect behaviour when not in live-results mode.


charisk (approved):

charisk:
Kind of unrelated to the PR, but are we planning to write some unit tests around query.ts? It's getting a bit more complex so it'd be good to have some coverage.

robertbrignull:
> Kind of unrelated to the PR, but are we planning to write some unit tests around query.ts? It's getting a bit more complex so it'd be good to have some coverage.

I would also love to have some tests of this code. I think we just need to do it, but it's hard to find the time, especially when there aren't many tests to start with. It may tie in to the work we're planning for mocking the github API in the extension and then we can use the same approach here. I'll at least add a note about this repo to the release confidence issue.

PR #746, opened 2022-09-21
Skip combine-results step if we have a variant analysis ID

robertbrignull:
When in live-results mode, we don't upload actions artifacts, but the `combine-results` job fails if there are no artifacts. But we don't actually need to run the `combine-results` job at all now, so it's purpose is now handled by the github API.

I think doing `always()` and another condition is valid syntax here. Weirdly it would have different behaviour to just the condition on it's own, because actions adds an implicit `if: success()` unless you use `always()` or `failure()`. See https://stackoverflow.com/questions/58858429 and https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions for more info.

We / I should ideally do some tests of this too and make sure it does behave as expected. If we can't get this condition to work, we could also just put the check in code at the start of `combine-results.ts`, but that's not ideal because then we'd start a whole actions runner just to exit immediately.

### DIFF
--- a/.github/workflows/codeql-query.yml
+++ b/.github/workflows/codeql-query.yml
@@ -96,7 +96,7 @@ jobs:
 
   combine-results:
     runs-on: ubuntu-latest
-    if: always()
+    if: ${{ always() && github.event.inputs.variant_analysis_id == '' }}
     needs:
       - run
     permissions: {}
--- a/.github/workflows/codeql-query.yml
+++ b/.github/workflows/codeql-query.yml
@@ -96,7 +96,7 @@ jobs:
 
   combine-results:
     runs-on: ubuntu-latest
-    if: ${{ always() && github.event.inputs.variant_analysis_id == '' }}
+    if: always() && github.event.inputs.variant_analysis_id == ''
     needs:
       - run
     permissions: {}


.github/workflows/codeql-query.yml
```diff
@@ -96,7 +96,7 @@ jobs:
 
   combine-results:
     runs-on: ubuntu-latest
-    if: always()
+    if: ${{ always() && github.event.inputs.variant_analysis_id == '' }}
```
koesie10, line 99: The `${{` shouldn't be necessary according to the [expressions page](https://docs.github.com/en/actions/learn-github-actions/expressions).

Have you tested whether this will evaluate correctly? For the `post-if` expression inside an `action.yml` file, this would actually not work because it's evaluating an expression inside an implicit expression and the `failure` function was undefined.

Another question: Are we able to use [the `inputs` context](https://docs.github.com/en/actions/learn-github-actions/contexts#inputs-context) instead of `github.event.inputs`?

robertbrignull: I've removed the `${{`. Thanks for pointing that out.

I don't know about the `inputs` context but I expect we just didn't know about it or didn't think too hard about it since `github.event.inputs` worked. I can see that it's a bit shorter though and that's nice. Just to keep changes separate I'll make that change in a followup PR.

I've also now tested this and it seems to be working correctly. Here are a few different runs:
- A run in live results mode - https://github.com/dsp-testing/qc-controller/actions/runs/3104378048
- A run in the old artifacts mode - https://github.com/dsp-testing/qc-controller/actions/runs/3104408456
- A run in the old artifacts mode where I cancelled it half way - https://github.com/dsp-testing/qc-controller/actions/runs/3104420815

koesie10 (approved):

PR #755, opened 2022-10-12
Add an integration test that triggers a variant analysis

robertbrignull:
Draft / test PR so far and then I'll fill this out properly later.


robertbrignull:
Should be good to review now. It seems to be all working now and it's adding the label successfully.

To check the non-pull request side of it, I did a run where it was running on `push` at https://github.com/github/codeql-variant-analysis-action/actions/runs/3282831674/jobs/5406762493. You can see that the variant analysis ran and the validation worked and the "success label" step was skipped successfully.

TIL actions prints out a very nice explanation of why it did or didn't run a step:
```
##[debug]Evaluating condition for step: 'Add success label to pull request'
##[debug]Evaluating: (success() && (github.event.pull_request.number != null))
##[debug]Evaluating And:
##[debug]..Evaluating success:
##[debug]..=> true
##[debug]..Evaluating NotEqual:
##[debug]....Evaluating Index:
##[debug]......Evaluating Index:
##[debug]........Evaluating Index:
##[debug]..........Evaluating github:
##[debug]..........=> Object
##[debug]..........Evaluating String:
##[debug]..........=> 'event'
##[debug]........=> Object
##[debug]........Evaluating String:
##[debug]........=> 'pull_request'
##[debug]......=> null
##[debug]....=> null
##[debug]....Evaluating Null:
##[debug]....=> null
##[debug]..=> false
##[debug]=> false
##[debug]Expanded: (true && (null != null))
##[debug]Result: false
```

.github/workflows/integration-test-call.yml
```diff
@@ -1,77 +0,0 @@
-name: "Integration test (using workflow call)"
```
koesie10, line 1: Do we not want to keep testing the existing MRVA flow? This is still used by our beta testers, so we still shouldn't break this until everyone is actually using live results.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,125 @@
...
+
+          ID="$(echo "$RESPONSE" | jq '.id')"
+          echo "Triggered variant analysis $ID"
+
+          if [ "$ID" == "null" ]; then
+            echo "Error triggering variant analysis"
+            exit 1
+          fi
+
+          echo "::set-output name=variant_anlaysis_id::$ID"
```
koesie10, line 51:
```suggestion
          echo "::set-output name=variant_analysis_id::$ID"
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: |
+          while true; do
+            RESPONSE=$(curl -H "Authorization: Bearer ${{ secrets.BOT_TOKEN }}" "https://api.github.com/repos/$GITHUB_REPOSITORY/code-scanning/codeql/variant-analyses/${{ steps.trigger.outputs.variant_anlaysis_id }}")
+            STATUS="$(echo "$RESPONSE" | jq '.status' -r)"
+            echo "Variant analysis ${{ steps.trigger.outputs.variant_anlaysis_id }} status: $STATUS"
+            if [ "$STATUS" == "completed" ]; then
+              echo "Exiting..."
+              exit 0
+            fi
+            sleep 10s
```
koesie10, line 63: Should we put a timeout on this workflow of e.g. 30 minutes? Otherwise we would keep polling for 6 hours.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,125 @@
...
+          done
+
+      - name: Validate variant anlaysis status
```
koesie10, line 66:
```suggestion
      - name: Validate variant analysis status
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,125 @@
...
+
+          if [ "$(echo "$RESPONSE" | jq '.skipped_repositories.no_codeql_db_repos.repository_count')" != "0" ]; then
+            echo "Number of no CodeQL DB skipped repos is not 0"
+            exit 1
+          fi
+
+          if [ "$(echo "$RESPONSE" | jq '.skipped_repositories.over_limit_repos.repository_count')" != "0" ]; then
+            echo "Number of over limit skipped repos is not 0"
+            exit 1
+          fi
```
koesie10, line 110: All of these tests seem to be testing the GitHub API, rather than this action. This action shouldn't influence any of these, right? The only thing this action should influence is the status of individual repo tasks, which doesn't seem to be checked here.

.github/workflows/integration-test-call.yml
robertbrignull: You're right we should keep testing it. I was a bit too hasty in trying to delete it.

I'd prefer not to compromise on the new approach too much trying to fit it in with the old approach, but that's a separate matter. If we're ok with having semi-duplicate checks on PRs for the next couple of months then that's fine.

.github/workflows/integration-tests.yml
robertbrignull: Good point. A timeout on the workflow would stop us wasting compute time, and if the variant analysis takes longer than (e.g.) 30 minutes then something is wrong and we can consider it to have failed.

.github/workflows/integration-tests.yml
robertbrignull: That's a good observation. I hadn't really considered it that way but you're right. I think this work has become a little conflated in my head with the idea of providing integration tests of all of variant analysis.

I'm now wondering if it makes sense to slim this test down to only consider the repo tasks, or whether to keep the current content and increase it further to cover the repo tasks. I'm leaning towards to second option as it's little extra effort and then it means these tests could be useful just to run on the `main` branch. Probably not possible / worth it for testing API changes before merging, but maybe just useful as a periodic sanity check.

koesie10: I agree that it would be really useful to have integration tests of the complete variant analysis flow, but I'm not sure that the correct place is within the action. Would it be possible to separate this out and think about implementing this as e.g. a scheduled workflow, independent of the action?

.github/workflows/integration-test-call.yml
robertbrignull: I've added the old tests back, or rather I've rebased this PR so it no longer removes the old tests. The first commit of this PR is the same as before but it no longer removes the old test code. Then the other commits after the first one are addressing comments or making changes.

.github/workflows/integration-tests.yml
robertbrignull: I've added some checks of the `scanned_repositories` section that checks that both analyses succeeded, have the correct number of results, and have an artifact uploaded.

Regarding the question of whether these tests belong here or test the right thing, I think the situation is not perfect but keeping the test here is best for now. At least I'd prefer to implement these new tests here and then think about moving them later. I don't think we need another set of integration tests, so I'd prefer to either keep them here or move them elsewhere but not have this style of tests in two places.

Honestly, running these tests on every pull request of the action seems unnecessary. The only reason it makes sense having them in this repository is because this is the only repo where it's possible to test changes in this way before you merge them, but these integration tests are testing the API at least as much as they are testing this action. So I wouldn't be opposed to moving these tests out of this action and converting them to only run on schedule and on demand. Then when making changes to this action you rely on the unit tests, and can run the integration tests manually if you deem it necessary.

Does that fit with how you're thinking, and how much of this do you think we need to answer before merging this PR?

koesie10: I think it would still make sense to run these integration tests as part of this action since this repository contains essentially all the code that runs as part of the variant analysis (including the workflow that is triggered and the actions that are run). I'm not sure there's a better place for these integration tests either since that would just be duplicating part of the integration tests from this repository.

I'm happy to merge the PR without having an answer on this question since it will definitely improve the state of the integration tests and we can think about having integration tests for all of variant analysis later.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,141 @@
...
+
+          ID="$(echo "$RESPONSE" | jq '.id')"
+          echo "Triggered variant analysis $ID"
+
+          if [ "$ID" == "null" ]; then
+            echo "Error triggering variant analysis"
+            exit 1
+          fi
+
+          echo "::set-output name=variant_analysis_id::$ID"
```
koesie10, line 52: Do we want to track the `set-output` deprecations as a separate issue? [The new method](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) seems to be to write to an environment file, but we will also need to modify the `codeql-query` workflow file for that.

koesie10 (approved):

robertbrignull:
Hmm, not sure what happened there but the CI wasn't triggering. I've squashed everything (you can see the force push is a no-op) and re-pushed and that seems to have got it going again.

This PR now only introduces the new workflow file and makes no changes to existing code. Since the new workflow was approved before I rebased I think this is good to go now and I'll merge it.

PR #757, opened 2022-10-21
Use local action instead of referencing branch

koesie10:
This will change the workflows to reference the workflow locally instead of referencing a branch. This will make it unnecessary to change the references in the workflow, making it easier to test changes to the action.

This also removes "integration tests succeeded" label since the GitHub Actions status will also be passing on the final commit. That means we should make the "integration-test" check required and remove the required state on "workflow-references-main".


robertbrignull (commented):
Ah this is really cool! I was under the impression that we couldn't reference local actions in these workflows because of how we triggered them via dynamic workflows. But if all it needed was the checkout the repository that's nice.

Have you tried out running a variant analysis with this branch?

koesie10:
> Have you tried out running a variant analysis with this branch?

I hadn't yet, but it seems like this won't work. The `actions/checkout` actions will actually check out the controller repository. We can easily change it to use this repository instead, but I don't think there's a way to figure out which ref the reusable workflow was called with. The easiest solution would be to add an input to the `codeql.query.yml` workflow with the `action_repo_ref` so we can check out the correct ref. I feel like this would be a relatively minor change which would provide some meaningful benefits for the DX of this action. What do you think?

robertbrignull:
> The easiest solution would be to add an input to the `codeql.query.yml` workflow with the `action_repo_ref` so we can check out the correct ref.

Hmm, that could work. I think it's worth trying out.

One thing to make clear is that you can't do templating in the `uses` value, so stuff like `- uses: actions/checkout@${{ github.event.inputs.ref }}` is not valid. I think this restriction is so it can evaluate these bits of the workflow at the start and then the system knows exactly which actions will be run.

But passing the ref as an argument to `actions/checkout` and then later on running local actions seems like it could work and solve our problems.

koesie10:
@robertbrignull This is ready for another review. Variant analyses now also work by passing in the `action_repo_ref` and remote queries will also automatically use the correct `action_repo_ref` when passing in a different one in the GitHub API request. This should allow us to get rid of the `script/use-this-branch`.

PR #759, opened 2022-10-25
Use `pull_request_target` event for triggering the integration test

koesie10:
The integration test is not being triggered for e.g. dependabot PRs. This changes the event to use `pull_request_target` to ensure that we have access to creating a status and triggering the workflow.

I'm not 100% sure about the implications of this, as it will change the security model of the integration test. However, it should only be possible to trigger this when you already have write access or when you have explicitly been approved (since the fork PR workflow permission is set to "Require approval for all outside collaborators").

### DIFF
--- a/.github/workflows/integration-test-trigger.yml
+++ b/.github/workflows/integration-test-trigger.yml
@@ -3,7 +3,11 @@ name: "Integration test (trigger only)"
 on:
   push:
     branches: [main]
-  pull_request:
+  pull_request_target:
+
+permissions:
+  actions: write
+  statuses: write
 
 jobs:
   trigger:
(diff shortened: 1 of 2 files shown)


koesie10:
It's hard to test since `pull_request_target` doesn't trigger on this PR, but I've been able to test it by rebasing #749 on top of this PR.

robertbrignull:
Tl;dr, I'm not 100% sure but I don't think this is what we want to do.

I'm getting my information from https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target and I think an important sentence from that page is:
> This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

This brings up two points:
- Because the PR runs on the base commit, does this mean we'd have to explicitly change the workflow to checkout the head or merge commit of the PR?
- We are using the `permissions` block here to limit the power of the github token, but I'm not sure if that's enough since we're also using repository secrets. We don't want to leak the bot account secret.

Another thing is, you mention "fork PR workflow permission" and it's good we have that set to always require approval. How does that apply to dependabot PRs, because those are not from forks? Another aspect is that we don't want to be vulnerable to rogue dependencies pushing unsafe code.

robertbrignull:
> The integration test is not being triggered for e.g. dependabot PRs.

Is it not getting triggered at all, or is it triggering but with read-only permissions and no secrets and therefore the workflows don't work?

PR #760, opened 2022-10-26
Replace deprecated `set-output` syntax with `GITHUB_OUTPUT`

shati-patel:
This PR should remove these deprecation warnings:
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

See internal issue too! 😊

### DIFF
--- a/.github/workflows/codeql-query.yml
+++ b/.github/workflows/codeql-query.yml
@@ -67,16 +67,16 @@ jobs:
         run: |
           const fs = require('fs');
           const process = require('process');
+
           const allRepos = JSON.parse(fs.readFileSync("instructions.json")).repositories;
           const repoNwos = new Set(${{ toJSON(matrix.repoNwos) }});
           const repositories = allRepos.filter(r => repoNwos.has(r.nwo));
           if ("${{ secrets.TEST_PAT }}") {
             allRepos.forEach(r => r.pat = "${{ secrets.TEST_PAT }}")
           }
+
           const githubOutput = process.env.GITHUB_OUTPUT;
-          const outputFile = fs.openSync(githubOutput, 'a');
-          fs.writeSync(outputFile, `repositories=${JSON.stringify(repositories)}`);
-          fs.closeSync(outputFile);
+          fs.appendFileSync(githubOutput, `repositories=${JSON.stringify(repositories)}`);
 
       # NOTE: this is only expected to work on github.com hosted runnners.
       # It will not work on any self-hosted runners.
--- a/.github/workflows/codeql-query.yml
+++ b/.github/workflows/codeql-query.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Checkout codeql-variant-analysis-action repository
         uses: actions/checkout@v3
         with:
-          repository: 'github/codeql-variant-analysis-action'
+          repository: "github/codeql-variant-analysis-action"
           ref: ${{ github.event.inputs.action_repo_ref }}
 
       - name: Mask download URL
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql-query.yml
```diff
@@ -66,13 +66,17 @@ jobs:
...
           const repoNwos = new Set(${{ toJSON(matrix.repoNwos) }});
           const repositories = allRepos.filter(r => repoNwos.has(r.nwo));
           if ("${{ secrets.TEST_PAT }}") {
             allRepos.forEach(r => r.pat = "${{ secrets.TEST_PAT }}")
           }
-          console.log(`::set-output name=repositories::${JSON.stringify(repositories)}`);
+          const githubOutput = process.env.GITHUB_OUTPUT;
+          const outputFile = fs.openSync(githubOutput, 'a');
+          fs.writeSync(outputFile, `repositories=${JSON.stringify(repositories)}`);
+          fs.closeSync(outputFile);
```
shati-patel, line 79: `@simon-engledew` helped me with this since this was a bit more complicated than the `>> $GITHUB_OUTPUT` version 😅 

We verified that this step succeeded in https://github.com/github/codeql-variant-analysis-action/actions/runs/3328681685/jobs/5504956701#step:5:111

image

koesie10, line 79: `@actions/core` seems to be using a single method for issuing these calls: https://github.com/actions/toolkit/blob/b36e70495fbee083eb20f600eafa9091d832577d/packages/core/src/file-command.ts#L22-L24. Can we also use `fs.appendFileSync` here instead of making these three separate calls (if the `require('@actions/core')` above doesn't work)?

robertbrignull:
```suggestion
          fs.appendFileSync(githubOutput, `repositories=${JSON.stringify(repositories)}`);
```
Could this been combined into one line by using [`appendFileSync`](https://nodejs.org/api/fs.html#fsappendfilesyncpath-data-options) or that not equivalent?

.github/workflows/codeql-query.yml
```diff
@@ -66,13 +66,17 @@ jobs:
         shell: node {0}
         run: |
           const fs = require('fs');
+          const process = require('process');
```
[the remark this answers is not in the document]
shati-patel: Thanks! Will give that a go 🥇

robertbrignull (approved):
One question but regardless this looks good to me

.github/workflows/codeql-query.yml
robertbrignull: Sorry my review was outdated. You can ignore it in favour of @koesie10's review.

.github/workflows/codeql-query.yml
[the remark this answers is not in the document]
shati-patel: [Didn't work](https://github.com/github/codeql-variant-analysis-action/actions/runs/3328859394/jobs/5505358697#step:5:110), sadly 😢

shati-patel:
Thanks for the reviews @koesie10 and @robertbrignull! 

Importing `@actions/core` didn't work (though we could probably make it work by using [`github-script`](https://github.com/actions/github-script) or something instead of a simple `node` shell), but I've made the `appendFileSync` change you suggested 🎉

.github/workflows/codeql-query.yml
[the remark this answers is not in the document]
koesie10: Thanks for trying it out!

PR #761, opened 2022-10-26
Use the correct SHA for status checks in PRs

koesie10:
In a PR, we should be using `${{ github.event.pull_request.head.sha }}` to ensure we are creating the status on the commit of the PR, rather than on the automatically created merge commit.

See: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request


shati-patel (approved):
Thanks for digging in to this 🙇🏽‍♀️ ✨

robertbrignull (approved):
Yep. Thanks for looking into it!

PR #762, opened 2022-10-27
Update individual repo tasks when overall workflow is canceled/failed

shati-patel:
Follow-up to #744. That PR added a `post` action that set repo tasks as "failed" when the overall workflow failed or was canceled.

This PR is an attempt to make that more fine-grained so that we can distinguish between the `failed` and `canceled` case. 

See internal issue for more details!


.github/workflows/codeql-query.yml
```diff
@@ -105,6 +105,28 @@ jobs:
           codeql: ${{ steps.init.outputs.codeql-path }}
           variant_analysis_id: ${{ github.event.inputs.variant_analysis_id }}
 
+      - name: Handle workflow canceled
+        if: canceled() && github.event.inputs.variant_analysis_id != ''
```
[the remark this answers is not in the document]
shati-patel: Thanks! Also fixed `failed()` -> `failure()` as per https://docs.github.com/en/actions/learn-github-actions/expressions#failure

robertbrignull:
Haven't done an extensive review but the rest of the PR looks good to me, so once the CI is passing I think this is looking pretty good ✨

shati-patel:
> Haven't done an extensive review but the rest of the PR looks good to me, so once the CI is passing I think this is looking pretty good ✨

Thanks! I found a bug while testing locally (see internal issue), so I'll have a poke around there first 👍🏽

src/update-repo-task-status.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!controllerRepoId || !variantAnalysisId || !workflowState) {
+    return;
+  }
+
+  const repos = getRepos();
+
+  for (const repo of repos) {
+    if (getState(`repo_${repo.id}_completed`)) {
+      continue;
+    }
```
shati-patel, line 32: Note to self:

It looks like the `saveState`/`getState` functions only work for post-actions, so they don't work here: 
image

Perhaps we could call the "get repo task" API here to check whether a repo task has been completed already.

shati-patel: (or we can pass the status along as outputs - see @robertbrignull's comment in the internal issue)

robertbrignull: Ooh, using the "get repo task" API is a good idea though. That way we'd be acting on more canonical data, rather than just what state changes we think we made already in the workflow. It would reduce the chance of weird edge-case bugs.

The only downside is it would increase the number of API request we make when a workflow is cancelled, but not by that much. I think it would add one request per repo, so we'd be doing 1-2 requests per repo instead of 0-1 requests. But I think we can handle this and it's not a problem. If we see scalability issues in the future we'll look at improving things, but right now I don't see any huge problems.

So right now I think we could go with either approach. Whatever you fancy implementing.

shati-patel:
Did some more local testing and confirmed that things work as expected (see internal issue!)

I believe this is now actually ready for review 😸

update-repo-task-status/action.yml
```diff
@@ -0,0 +1,22 @@
+name: "CodeQL: Update repository task status"
+description: "Set repository task as failed or canceled"
+author: "GitHub"
+inputs:
+  workflow_status:
+    description: "The status of the workflow: 'canceled' or 'failed'."
```
koesie10, line 6: Can we also mark this as `required: true`?

src/update-repo-task-status.ts
```diff
@@ -0,0 +1,72 @@
...
+import {
+  getControllerRepoId,
+  getRepos,
+  getVariantAnalysisId,
+  getWorkflowStatus,
+} from "./inputs";
+
+// All states that indicate the repository has been scanned and cannot
+// change status anymore.
+const COMPLETED_STATES = ["succeeded", "failed", "canceled", "timed_out"];
```
koesie10, line 16: Do you think this would be more resilient to change if we turn this around and list all statuses in which the variant analysis is still incomplete (i.e. `pending` and `in_progress`)? If we were to add a status, I would expect it to be a final status (for example another error) rather than a non-final status.

src/update-repo-task-status.ts
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * If the overall variant analysis workflow failed or was canceled,
+ * propagate the failure/cancellation status to the individual repo tasks.
+ */
+async function updateRepoTaskStatus(): Promise<void> {
+  const controllerRepoId = getControllerRepoId();
+  const variantAnalysisId = getVariantAnalysisId();
+  const workflowStatus = getWorkflowStatus();
+
+  if (!controllerRepoId || !variantAnalysisId || !workflowStatus) {
```
koesie10, line 27: I don't think this is actually possible: if any of these are not passed, the `getInput` function should already throw. The types also don't specify that these can be `undefined`.

koesie10 (approved):

src/update-repo-task-status.ts
shati-patel: Yes! Good point, thanks ✨

src/update-repo-task-status.ts
shati-patel: I think you're right ✔️  Removed that block!

PR #763, opened 2022-10-28
Fix BQRS files being corrupted when zipped

koesie10:
See the two commits:
- We do not decode the read file into a string anymore, so this should fix the corruption of the BQRS files
- We now store the files in the ZIP using `DEFLATE` compression instead of using the default `STORE` compression. This should reduce the uploaded file size since the files are now actually compressed


robertbrignull (approved):
LGTM

I tried it out with a non-problem query and the results can now be shown correctly.

I also checked the artifact size. For my query it reduced it from 116920 to 17638 💪

robertbrignull:
> Sorry a bit of a drive by comment rather than a review: do we need to keep the existing code as is so that the existing ("remote queries") flow doesn't break?

I just tried it out in "remote queries" mode too and it still works when using this branch, so I think it's all fine.

PR #767, opened 2022-11-09
Handle early cancellation of workflow runs

koesie10:
This will add a new job to the end of the workflow which will only run when the variant analysis is cancelled. It will update any repo tasks which do not have a final state set to be set to the `canceled`/`failed` status.

We need to use two separate jobs because the `cancelled()` and `failure()` return a different result based on where they are called. If they are called from within a step, they will return whether the current job has been cancelled or has failed, whereas if they are on the job it will be on the workflow. We may be able to create a composite action to reduce the duplication, but this seems like it can be a future improvement.

This can be optimized by using the variant analysis summary endpoint, but I don't believe we have access to that endpoint when using the signed auth tokens from #766.


.github/workflows/codeql-query.yml
```diff
@@ -165,3 +165,93 @@ jobs:
...
+
+    steps:
+      # To use the action in this repository, we need to check out the repository
+      - name: Checkout codeql-variant-analysis-action repository
+        uses: actions/checkout@v3
+        with:
+          repository: "github/codeql-variant-analysis-action"
+          ref: ${{ github.event.inputs.action_repo_ref }}
+
+      - name: Mask download URL
```
robertbrignull, line 186: We should mask the signed URL in these jobs too. See https://github.com/github/codeql-variant-analysis-action/blob/84ca3c82f1a66d4f8409ff3ff95881299c401ee5/.github/workflows/codeql-query.yml#L49

robertbrignull (commented):
The actual code looks fine, with one addition.

You mentioned elsewhere that we could remove the `if: cancelled()` step but need to keep the `if: failed()` one. Shall we do that in this PR?

I think the concerns about rate limiting are also valid. I'm not fully sure what we should do here but I think this PR can go ahead and we can revert it if needed. The fallout should be limited to just us developing live results.

What do you think?

koesie10:
@robertbrignull I've removed the cancellation step from the "run" job, so it will now all be handled by the extra cancellation job. I think we can go ahead with merging this PR, the rate limits shouldn't kick in until you're running multiple variant analyses with 1,000 repositories per hour which doesn't happen that often while developing live results.

.github/workflows/codeql-query.yml
```diff
@@ -165,3 +154,103 @@ jobs:
...
+          echo "repositories=$(jq --raw-output --compact-output .repositories instructions.json)" >> $GITHUB_OUTPUT
+
+      - name: Handle workflow cancelled
+        uses: ./update-repo-task-status
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          workflow_status: "canceled"
+          controller_repo_id: ${{ github.repository_id }}
+          repositories: ${{ steps.repos.outputs.repositories }}
```
robertbrignull, line 205: I tried to run this on a variant analysis of 1000 repos and this step failed with
```
An error occurred trying to start process '/home/runner/runners/2.299.1/externals/node16/bin/node' with working directory '/home/runner/work/qc-controller/qc-controller'. Argument list too long
```
I think because it's passing the entire repositories input on the command line to the `node` process.

I think we can work around this by having the `update-repo-task-status` action read the inputs directly from the `$GITHUB_EVENT_PATH` file, rather than accepting them from the `with` block in the workflow file.

koesie10: Thanks for noticing this, I've updated the action. However, instead of reading the inputs from the `GITHUB_EVENT_PATH`, we need to read the repositories from the instructions file (to which we only have a URL in the `GITHUB_EVENT_PATH` file). So, the workflow will still download the instructions file and pass the download path to the action, which will read the file and extract the `repositories` from it.

robertbrignull:
The code looks good to me and I've tried it out and it seems to be working successfully now. The job to set 1000 repo tasks to `cancelled` took over 8 minutes, but I think that's fine for now and there's not much we can do right now, and we already know the potential improvements we want to make in the future.

Did the integration tests pass successfully? It looks like they timed out?

PR #771, opened 2022-11-14
Use new variant analysis statuses in integration tests

koesie10:
The variant analysis status enum has been updated. This will change the integration tests to use the correct new status.


robertbrignull (approved):
👍

PR #775, opened 2022-11-24
Add JSON schemas for more types

robertbrignull:
Following on from https://github.com/github/codeql-variant-analysis-action/pull/773, this PR introduces more JSON schemas and add checking to every remaining place we do `JSON.parse`. This includes:
- Sarif parsing
- Query run metadata
- The output various CodeQL commands:
  - `codeql bqrs info`
  - `codeql resolve database`
  - `codeql resolve queries`

With this we cover every place where we do `JSON.parse`. This will allow us to be much more type-safe and confident in our types, by catching any unexpected data as early as possible.

In preparation I switched the exports from `codeql.ts` to being an `export` on each type/method rather than a big `export` statement. It seemed that `ts-json-schema-generator` wasn't understanding it and couldn't find the types.

Also as an observation, running `generate-json-schema` is getting quite slow and now takes about 30 seconds. I assume it's because it starts up a node process for each one. Hopefully it won't be a huge issue since you'll only run it yourself infrequently when you add a new schema. If it is a problem we can look at making it faster.


src/codeql.ts
```diff
@@ -189,7 +189,10 @@ export async function getBqrsInfo(
       `Unable to run codeql bqrs info. Exit code: ${bqrsInfoOutput.exitCode}`
     );
   }
-  return deserialize(bqrsInfoOutput.stdout);
+  return validateObject(
```
koesie10, line 192: Will this message be confusing to users/us? The message might say something like "missing `compatibleQueryKinds`", while the actual property in the file could be called `compatible_query_kinds`. Would it make sense to remove the camelization in a future PR?

koesie10 (approved):
I think we can improve the speed at which the JSON schema generation is running. To speed it up significantly by using a script, we can add `--no-type-check` to the call to `ts-json-schema-generator`. On my machine, this speeds it up from 37s to 13s.

If we want to further improve the speed, we can create a Node script that calls `ts-json-schema-generator` which will significantly improve the speed by re-using the same generator. I used this script to test this idea:

<details>
<summary><code>src/generate-json-schemas.ts</code></summary>

```ts
import * as fs from "fs";
import * as path from "path";

import * as tsj from "ts-json-schema-generator";

const rootDirectory = path.resolve(__dirname, "..");

const schemas = [
  {
    path: path.resolve(rootDirectory, "src/inputs.ts"),
    types: ["RepoArray", "Instructions"],
  },
  {
    path: path.resolve(rootDirectory, "src/codeql.ts"),
    types: [
      "Sarif",
      "BQRSInfo",
      "ResolvedQueries",
      "ResolvedDatabase",
      "QueryRunMetadata",
    ],
  },
  {
    path: path.resolve(rootDirectory, "src/query-run-metadata.ts"),
    types: ["QueryRunMetadata"],
  },
];

try {
  fs.rmdirSync(path.resolve(rootDirectory, "json-schemas-tmp"), {
    recursive: true,
  });
} catch (e) {
  // Ignore
}
fs.mkdirSync(path.resolve(rootDirectory, "json-schemas-tmp"));

for (const schema of schemas) {
  const generator = tsj.createGenerator({
    path: schema.path,
    tsconfig: path.resolve(rootDirectory, "tsconfig.json"),
    skipTypeCheck: true,
    topRef: true,
    additionalProperties: true,
  });

  for (const type of schema.types) {
    const generatedSchema = generator.createSchema(type);
    const schemaPath = path.resolve(
      rootDirectory,
      "json-schemas-tmp",
      `${type}.json`
    );
    fs.writeFileSync(schemaPath, JSON.stringify(generatedSchema, null, 2));
  }
}
```
</details>

We don't have any good options for running/compiling scripts in the extension right now and probably don't want the overhead of including `ts-node`, so I just placed it in `src/generate-json-schemas.ts` and called it using `node lib/generate-json-schemas.js`. When I run this, it takes about 3 seconds to generate all schemas. The script is still missing the code to move the files to the correct location, but that time should be negligible.

I don't think the script approach is worth it right now since it would pollute the `src` directory, but adding `--no-type-check` does seem worth it.

robertbrignull:
I've added `--no-type-checks` and this brings the runtime down from 40s to 17s. I think this is good enough to merge this PR.

I'll open issues to look into the script to make it faster, and about the camelization.

PR #776, opened 2022-11-24
Remove deprecated code

shati-patel:
🧹 Cleans up code that doesn't get used:

- We no longer need `security-events: write` permission for the MRVA endpoints
- We no longer have a `token` field in the `repositories` object
  - Ideally, we'd also get rid of `pat`, since that only [gets used in old integration tests](https://github.com/github/codeql-variant-analysis-action/blob/2e07ba3fbddcede187f3b53f552c425189b76f83/.github/workflows/codeql-query.yml#L80-L82). We can remove that once live results is fully available.

See internal linked issue for details!

### DIFF
--- a/.github/workflows/codeql-query.yml
+++ b/.github/workflows/codeql-query.yml
@@ -34,7 +34,6 @@ jobs:
       matrix:
         repoNwos: ${{ fromJSON(needs.setup.outputs.repoNwoChunks) }}
     permissions:
-      security-events: write # This permission can be removed soon
       contents: write
 
     steps:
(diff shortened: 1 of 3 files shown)


robertbrignull (approved):
Integration tests pass and my test query with this branch was successful, so LGTM

robertbrignull:
Ok, actually, I just found a couple more references, but only in comments. If you could remove those in this PR too that would be great. See
- https://github.com/github/codeql-variant-analysis-action/blob/main/query/action.yml#L18
- https://github.com/github/codeql-variant-analysis-action/blob/main/update-repo-task-status/action.yml#L14

shati-patel:
> Ok, actually, I just found a couple more references, but only in comments. If you could remove those in this PR too that would be great. See
> 
> * https://github.com/github/codeql-variant-analysis-action/blob/main/query/action.yml#L18
> * https://github.com/github/codeql-variant-analysis-action/blob/main/update-repo-task-status/action.yml#L14

Oooh, thank you, good spot!! 🦅 👀  Updated.

PR #777, opened 2022-11-25
Enable the remaining eslint rules

robertbrignull:
Using the new JSON schemas, we can now enable all of the remaining ESLint rules that we want. These rules were disabled during the transition from tslint to eslint because we didn't have time to fix them at the time of the upgrade.

See each individual commit for the changes made to fix that rule. The general theme of the changes is avoiding using `any` and being sure about the type of an object before we access anything on it.

Sometimes this does make the code a bit more complex. I don't think all of these changes are an improvement in all ways. I'm very happy to discuss the merits of any of the changes. We don't necessarily have to enable all of these rules. The combination of all of them does make it a little hard to work at times.

I have chosen to leave the `@typescript-eslint/restrict-template-expressions` rule disabled. This seems too difficult to accommodate and also not to offer as much benefit as the others.


src/deserialize.ts
```diff
@@ -1,7 +1,8 @@
-export function camelize(_: string, value: any) {
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
koesie10, line 1: Can we just disable this rule on the line where it's actually used, rather than on the complete file, using something like `// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment`?

.eslintrc.json
```diff
@@ -66,16 +66,8 @@
...
             "files": "**",
             "rules": {
-                "@typescript-eslint/explicit-module-boundary-types": "off",
-                "@typescript-eslint/no-explicit-any": "off",
-                "@typescript-eslint/no-unsafe-assignment": "off",
-                "@typescript-eslint/no-unsafe-call": "off",
-                "@typescript-eslint/no-unsafe-member-access": "off",
-                "@typescript-eslint/no-unsafe-return": "off",
-                "@typescript-eslint/prefer-regexp-exec": "off",
                 "@typescript-eslint/restrict-template-expressions": "off"
```
koesie10, line 71: Since we're leaving this disabled, can we move it to the `rules` configuration option, rather than leaving it in an override for all files?

koesie10 (approved):

src/deserialize.ts
robertbrignull: I'd have to update `.eslintrc` because currently we say that the only eslint comments that are allowed are `eslint-disable` and `eslint-enable`, so that doesn't include `eslint-disable-next-line`. But from reading https://github.com/github/codeql-variant-analysis-action/pull/739 I don't think it was intended to disallow `eslint-disable-next-line` but rather in that PR we didn't need it.

robertbrignull:
I've done some test runs using this branch of a successful analysis, an analysis of a non-problem query, and a cancelled analysis. It's all gone smoothly, so I think this should be fine to merge and it won't break anything. I'm most worried about the new json schemas catching a case where the json is currently invalid but we haven't noticed, but I think the above tests should have caught that.

PR #778, opened 2022-11-25
Install `node_modules` on Windows

shati-patel:
🚧 WIP 🚧


shati-patel:
That definitely didn't work 🤓

PR #782, opened 2022-12-06
Do not allow reruns of the workflow

koesie10:
Example run with re-running only failed jobs: https://github.com/github/codeql-variant-analysis-action/actions/runs/3628283507/attempts/2

Example run with re-running setup job: https://github.com/github/codeql-variant-analysis-action/actions/runs/3628283507/attempts/3


charisk (approved):
LGTM

PR #783, opened 2022-12-06
Add error messages when downloading instructions or database fails

koesie10:
This adds an error message that explains that the instructions file or database download can fail because it expires after 24 hours.


charisk (approved):
LGTM!

PR #790, opened 2023-01-12
Use new update endpoint

koesie10:
This will update the endpoint which is used to update the status of a variant analysis repo task.


norascheuch (approved):
🚀

PR #791, opened 2023-01-13
Remove unused function for getting repo task

koesie10:
In the Action, we're not actually retrieving the repo task and this could be misleading us into thinking that we're retrieving the repo task.


robertbrignull (approved):
Good spot!

I guess this became unused when we changed to making a single request to update all repo tasks in case of failure or cancellation.

PR #801, opened 2023-02-14
Delete old code that is not used in the live-results flow

robertbrignull:
This code should all be unused now. The GitHub API method that triggered this code path has also now been removed.

I've done a couple of MRVA runs using this branch and it seems to be working just fine.


.github/workflows/codeql-query.yml
```diff
@@ -135,7 +135,7 @@ jobs:
           signed_auth_token: ${{ github.event.inputs.signed_auth_token }}
 
       - name: Handle workflow failed
-        if: failure() && github.event.inputs.variant_analysis_id != '' && github.run_attempt == 1
```
koesie10, line 138: I think we can also make the `variant_analysis_id` required in the `query/action.yml` file, but this can perhaps be done in a different PR since I believe that the `signed_auth_token` should also be required now.

koesie10 (approved):
LGTM, I also did a MRVA run using this branch and couldn't find any issues with it

.github/workflows/codeql-query.yml
robertbrignull: Yeah I agree there's some signed auth token changes we can make too. Since we're happy with this as it is I'll merge this now and open a new PR to make both of the fields required.

robertbrignull:
In order to merge this, I removed the branch protection requirement for the `integration-test` CI status that was removed in this PR.

.github/workflows/codeql-query.yml
robertbrignull: Followup PR: https://github.com/github/codeql-variant-analysis-action/pull/802

PR #814, opened 2023-04-25
Retry uploading artifact when it fails with a 504 response

robertbrignull:
This PR aims to reduce the cases of result artifact upload failing because of a 504 response code. From anecdotal experience, this happens quite a lot. Sometimes multiple times in a 1000 repository MRVA run.

I have experimentally tested what happens when we re-upload the same artifact after an error, and the behaviour is that it overwrites rather than appending to any existing content. Therefore it should be safe to retry on errors like this, but to limit the risk I've kept the retrying behaviour to just a 504 response.

The first commit is just pulling code out from `download.ts` into separate files so it is easier to re-use. Then the second commits adapts `uploadArtifact` to use `RetryHelper`.


robertbrignull:
I think the CI failures are because the node version on the actions runner has updated enough to cause issues. We haven't noticed because we haven't updated this repo in so long. I'll look into it tomorrow and open a separate PR to fix it.

charisk (approved):
LGTM! Thank you for the two separate commits - made it much easier to review.

PR #815, opened 2023-04-27
Fix CI after node/npm versions have changed

robertbrignull:
This PR aims to get the CI working again. I think what's happened is that the version of Node/NPM used on actions runners has increased. We now need to upgrade our lock file in order to get CI to pass again. I'm also hitting issues when generating json schemas.

I generated the package lock change by running `npm ci && npm run removeNPMAbsolutePaths` from a codespace.

The node/npm versions are:
| | codespace | actions |
|---|---|---|
| Node | 19.8.1 | 18.16.0 |
| NPM | 9.5.1 | 9.5.1 |

### DIFF
--- a/node_modules/.package-lock.json
+++ b/node_modules/.package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "codeql-query",
   "version": "1.0.7",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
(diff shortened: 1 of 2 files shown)


charisk (commented):
Seems sensible to me! Thanks for fixing.

charisk (approved):

PR #820, opened 2023-05-11
Remove required permissions from jobs

charisk:
These are no longer required since we're using signed auth tokens.

### DIFF
--- a/.github/workflows/codeql-query.yml
+++ b/.github/workflows/codeql-query.yml
@@ -46,8 +46,7 @@ jobs:
       fail-fast: false
       matrix:
         repoNwos: ${{ fromJSON(needs.setup.outputs.repoNwoChunks) }}
-    permissions:
-      contents: write
+    permissions: {}
 
     steps:
       - name: Do not allow reruns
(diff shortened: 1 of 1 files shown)


robertbrignull:
There are a few places in `.github/workflows/codeql-query.yml` that we reference `GITHUB_TOKEN`, and I think all of those can be removed. We are passing the token into the action as an env var but then we aren't actually using the token in the action, because all requests are now authenticated using the signed auth token.

I suggest we try deleting all references to `GITHUB_TOKEN` and then do another manual test to confirm it still works.

robertbrignull:
(As said on slack as well), the general approach here looks good to me. We aren't using these permissions anymore so the right thing to do is to remove the requirement for them.

.github/workflows/codeql-query.yml
```diff
@@ -122,8 +121,6 @@ jobs:
 
       - name: Run query
         uses: ./query
-        env:
```
robertbrignull, line 125: Ok, unfortunately it does not work. See https://github.com/dsp-testing/qc-controller/actions/runs/4981941371

Seems it's the `@octokit/auth-action` dependency requiring that, even though we end up not using it and overwriting the authorization header with the signed auth token. We might be able to hack around that, but it's not essential and probably not worth it in this PR. What do you think?

charisk: Thanks for looking at that before I got to it - saved me some time. 

Yeah I think we can take that as tech debt for now.

robertbrignull (approved):

PR #827, opened 2023-06-01
Make it possible to specify runner os by setting a repo actions variable

norascheuch:


robertbrignull (approved):
LGTM

Works for me when the actions variable is not set, so shouldn't affect any users not using the variable.

PR #845, opened 2023-08-15
Switch YAML parser to `yaml` to fix SHA parsing bug

koesie10:
This switches the YAML parser from `js-yaml` to `yaml` to fix a bug when parsing a `database_commit_sha` which matches the exponential notation. `yaml` allows for more customization than `js-yaml`, so it allows us to override the regex that is used for identifying exponential floats.

The fix is changing the regex to make the use of `-` or `+` mandatory, such that `4225332178759948e04347560002921719079454` is not a valid float, but `4225332178759948e+04347560002921719079454` would be. This matches the regex in [the YAML 1.1 specification](https://yaml.org/type/float.html) and in [SnakeYAML](https://bitbucket.org/snakeyaml/snakeyaml/src/180363cfff5c9334bdd21b0d38145fa56ceed91c/src/main/java/org/yaml/snakeyaml/resolver/Resolver.java#lines-37).

This also fixes it when the complete commit SHA only contains digits, but I don't think the CodeQL CLI would output a non-quoted value in that case. Nevertheless, this fixes it by parsing all ints as `BigInt` and calling `toString()` on the value.

I've split the changes into separate commits, so I'd recommend to review this commit-by-commit.


src/yaml.ts
```diff
@@ -0,0 +1,46 @@
...
+      return num.toExponential();
+    }
+
+    return isNaN(num) ? ".nan" : num < 0 ? "-.inf" : ".inf";
+  },
+};
+
+export function parseYaml<T>(src: string): T {
+  return parse(src, {
+    version: "1.1", // CodeQL CLI uses YAML 1.1
```
robertbrignull, line 26: Just to confirm, specifying `version: "1.1"` here isn't enough to make the library use the regex from the YAML 1.1 specification? What exactly does this version value control?

koesie10: Unfortunately not, the library [does not use](https://github.com/eemeli/yaml/blob/07063bc865b862c2dc375630e27078b2886d4668/src/schema/yaml-1.1/float.ts#L24) the exact regex from the YAML 1.1 specification, so we need to override it. There do seem to be some changes between [the regex used for YAML 1.2 in the library](https://github.com/eemeli/yaml/blob/07063bc865b862c2dc375630e27078b2886d4668/src/schema/core/float.ts#L24), so there are some differences in parsing.

robertbrignull (approved):
Sorry for the delay in reviewing this. I'm finding this problem area a little confusing, so I'm not totally sure what the best solution is. But merging this change is likely fine, and it's also easily reversible if we decide to do something else in the future. Let's go ahead with this.

PR #847, opened 2023-08-15
Update types to node16

charisk:
We're using node16 in the action so this PR updates the types to use that version.


charisk:
Thanks for catching that - my code search failed me!

robertbrignull:
I'm sorry too as I gave some unhelpful advice earlier and also forgot the node version is specified in the `action.yml` rather than the workflow.

koesie10 (approved):
Thanks, these changes look good to me

PR #853, opened 2023-08-22
Only upload results if there are any

charisk:
This will need to be merged after https://github.com/github/vscode-codeql/pull/2736 has shipped (has been included in a released version of the extension)


robertbrignull:
Code change LGTM but (as posted on internal issue) we might need to look at the API methods and make sure they'll accept us calling them without uploading the artefact.

charisk:
Abandoning this change for now until we make some changes in the monolith that will help us have more correct data.

PR #854, opened 2023-08-23
Always include BQRS in results archive

shati-patel:
Instead of choosing whether to include SARIF or BQRS results in the results download, we always include at least BQRS. This is so that users have to option to display raw results in the MRVA results view, even if SARIF is available. 

See internal linked issue for more details!


shati-patel:
Thanks! I've tested with this branch and everything is conveniently handled (or ignored) on the VS Code side 😅

PR #861, opened 2023-09-15
Upgrade Ava to 5.3.1

koesie10:
This upgrades the ava package to version 5.3.1 and upgrades `@ava/typescript` to `4.1.0` to keep in sync. There were some changes in the tests necessary due to some breaking changes in the names of the TypeScript interfaces and the return type of the `t.throws` method. I also had to disable `workerThreads` because we're using `process.chdir` and this is not supported in worker threads.

I was getting a linting error because `eslint-plugin-import` was using a resolver which does not support Ava's `exports` in its `package.json`. I used `eslint-import-resolver-typescript` to use TypeScript's resolver instead (since this is also used for building).


robertbrignull (commented):
LGTM but has a lot of conflicts now

Thanks for working out how to get this upgraded and working

robertbrignull (approved):

PR #862, opened 2023-09-15
Use bundled files using ESBuild rather than checking in `node_modules`

koesie10:
This PR implements checking in 3 generated JavaScript files rather than checking in both all generated JavaScript files and the `node_modules` directory. Due to the much fewer number of files, this should be faster to check out and should also be easier to deal with (no discrepancies between operating systems such as `fsevents`).

This matches how standard actions such as actions/checkout and actions/setup-node deal with dependencies, although they are using ncc instead of ESBuild.

The most obvious disadvantage is that we're now running tests from a different set of files than what we ship to users. The tests are still run from the generated `lib` files and dependencies are imported from `node_modules`. I don't think this is a big problem since they should behave identically.


robertbrignull (approved):
LGTM, though you may want to get consensus from the rest of the team.

I'm happy with this approach. It is true that it opens up to a class of bugs where the behaves differently when built compared to when running "locally", but you would hope that such bugs are pretty rare. I think the benefit of not having the node_modules directory checked into the repo outweighs that. This will make it far easier for us to add or upgrade dependencies, and make using dependabot far easier.

I also like that we're copying the approach used by repos from the actions org. Copying them is probably a safe bet.

charisk:
I'm happy with the approach personally, specially since it's used elsewhere too. I've not reviewed the actual code though or tested this.

koesie10:
> LGTM, though you may want to get consensus from the rest of the team.

I've asked `@starcke` and he doesn't have any objections to this and since `@charisk` has already said she's happy with the approach, I'll merge this since the others are OOF for at least a few days more. We can always revert this if necessary.

PR #863, opened 2023-09-18
Upgrade TypeScript and ESLint

koesie10:
To upgrade TypeScript, we also need to upgrade ESLint for compatibility with this version of TS.

This PR supersedes #855


robertbrignull (approved):
LGTM and works for me

PR #878, opened 2023-09-26
Use original query metadata

koesie10:
This changes the `codeql bqrs interpret` call to use the original query metadata rather than passing in only a fixed `-t=id=remote-query` argument. This should make MRVA behave more like running a local query since this is [what we also do in the VS Code extension](https://github.com/github/vscode-codeql/blob/31e233cc59c71c664b39d39d40b40f03e13e1737/extensions/ql-vscode/src/codeql-cli/cli.ts#L1057).

We still need to ensure that there is an ID passed in since interpreting a SARIF file requires an ID, so we still have a fallback if we do execute a query without an ID (I'm not sure if that's possible though).

### DIFF
--- a/dist/query.js
+++ b/dist/query.js
@@ -61587,6 +61587,9 @@ var QueryMetadata_default = {
   definitions: {
     QueryMetadata: {
       properties: {
+        id: {
+          type: "string"
+        },
         kind: {
           type: "string"
         }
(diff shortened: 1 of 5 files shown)


shati-patel (approved):
Nice, looks good to me 🏆

PR #889, opened 2023-10-11
Fix support for `codeql-pack.yml` file

koesie10:
When the `qlpack.yml` file didn't exist, `fs.statSync` would throw an error. Therefore, we would never check if the `codeql-pack.yml` exists.

This fixes it by using the `throwIfNoEntry` option of `fs.statSync`.


charisk (approved):
LGTM!

PR #895, opened 2023-10-18
Make variant analysis workflow so it can be executed directly as a dynamic workflow

robertbrignull:
The idea here is to make it so that the GitHub API can read and execute this workflow file directly, instead of calling it as a callable workflow.

The aim is to move the file out of the `.github/workflows` directory because it isn't really a workflow that's executed in the normal way. It's just a YAML file that happens to contain workflow code. For now, this PR duplicates the file and leaves the old one alone. This will make it easier to deploy and roll back if needed, rather than having to merge this at the same time as the API change.

See the internal github PR for the other side of this change.

### DIFF
--- a/variant-analysis-workflow.yml
+++ b/variant-analysis-workflow.yml
@@ -1,0 +1,241 @@
+name: ${{ github.event.inputs.workflow_name }}
+
+env:
+  CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT: true
+
+on: dynamic
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      repoNwoChunks: ${{ steps.split.outputs.repoNwoChunks }}
+    permissions: {}
+
+    steps:
+      - name: Do not allow reruns
+        if: github.run_attempt != 1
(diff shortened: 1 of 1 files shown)


variant-analysis-workflow.yml
```diff
@@ -0,0 +1,241 @@
+name: ${{ github.event.inputs.workflow_name }}
+
+env:
+  CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT: true
+
+on: dynamic
```
charisk, line 6: Is this missing the `TEST_PAT` stuff intentionally?  How do you see tests working once the other workflow file is gone?

charisk (approved):
LGTM

variant-analysis-workflow.yml
robertbrignull: Thanks for raising this. I think you've pointed out some tech debt and the `TEST_PAT` is actually not used, maybe? I certainly can't find any place where it is set from, so I think it's always empty and unused.

Unless we find out where it's set from, I'm going to remove references to it from the new workflow file.

PR #901, opened 2023-10-25
Use run-name which can accept dynamic expressions

robertbrignull:
This fixes MRVA when using the new feature of reading the dynamic workflow directly from this repo. Unfortunately the `name` field cannot use expressions, so workflow runs were failing. Instead we can use `run-name` which does work with expressions.

Currently this feature is only enabled on my private test controller repo, but I have screenshots of a successful run I did using this branch.

Screenshot 2023-10-25 at 11 21 11
Screenshot 2023-10-25 at 11 21 21

### DIFF
--- a/variant-analysis-workflow.yml
+++ b/variant-analysis-workflow.yml
@@ -1,4 +1,5 @@
 name: CodeQL Variant Analysis
+# `run-name` can contain expressions, and will override the `name` once the workflow starts running
 run-name: ${{ github.event.inputs.workflow_name }}
 
 env:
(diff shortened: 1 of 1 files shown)


charisk (approved):

variant-analysis-workflow.yml
```diff
@@ -1,4 +1,5 @@
-name: ${{ github.event.inputs.workflow_name }}
+name: CodeQL Variant Analysis
+run-name: ${{ github.event.inputs.workflow_name }}
```
charisk, line 2: Might be worth adding a comment to explain why we use run-name

robertbrignull, line 2:
```suggestion
# `run-name` can contain expressions, and will override the `name` once the workflow starts running
run-name: ${{ github.event.inputs.workflow_name }}
```

PR #956, opened 2024-01-09
Remove unused file size fields from RunQueryResult

charisk:
Remove `bqrsFileSize` and `sarifFileSize` from `RunQueryResult` as they don't actually seem to be used anywhere.


shati-patel (approved):
Looks good, thanks! 🧹

PR #957, opened 2024-01-09
Remove unused query run metadata

charisk:
It doesn't seem we're using the query run metadata any more, so I've removed it.


shati-patel (approved):
Thanks for the cleanup! We clearly haven't been doing a good enough job of tidying up our code 😅

PR #958, opened 2024-01-09
Add running tests section in CONTRIBUTING guide

charisk:
Added a call out to `script/test` that I wasn't aware of before @shati-patel mentioned it to me.


shati-patel (approved):
✨

PR #959, opened 2024-01-09
Add tsbuildinfo file to gitignore

charisk:
The `tsconfig.tsbuildinfo` file has build information file generated by TypeScript, to help detect what has changed between compilations, allowing it to compile only the changed parts and thus speeding up the compilation process.

I don't think we should be checking this in, and since it's generated, I think it should be in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /lib/
 /node_modules/
 /dist/*.js.map
+tsconfig.tsbuildinfo


robertbrignull (approved):
I don't seem to have one of these files locally. But I have no objection to ignoring it.

charisk:
> I don't seem to have one of these files locally. But I have no objection to ignoring it.

I get it after running `npm run watch`

PR #961, opened 2024-01-09
Add multi-query support

charisk:
Adds initial support for running multiple queries by making sure that multiple BQRS files are allowed and processed.

This is heavily inspired by the spike that @robertbrignull and I did a few weeks ago (see internal issue for details).


src/codeql.ts
```diff
@@ -230,6 +277,102 @@ async function getSourceLocationPrefix(codeql: string) {
...
+  queries: Array<{
+    queryPath: string;
+    relativeBqrsFilePath: string;
+    bqrsInfo: BQRSInfo;
+  }>;
+  totalResultsCount: number;
+  resultsBasePath: string;
+}
+
+async function getQueryPackRunResults(
```
robertbrignull, line 290: Minor point but I find some of the names a bit of a mouthful. Do they need to have `QueryPack` in the name or would `getRunResults` or `getQueryResults` be ok?

src/codeql.ts
```diff
@@ -377,19 +501,19 @@ export function getDatabaseMetadata(database: string): DatabaseMetadata {
...
 /**
- * Gets the query for a pack, assuming there is a single query in that pack's default suite.
+ * Gets the queries for a pack.
  *
  * @param codeql The path to the codeql CLI
  * @param queryPack The path to the query pack on disk.
  * @returns The path to a query file.
  */
-export async function getRemoteQueryPackDefaultQuery(
+export async function getRemoteQueryPackQueries(
```
robertbrignull, line 513: Does it need `Remote` in the name, or could this be `getQueryPackQueries`?

src/codeql.ts
```diff
@@ -230,6 +277,102 @@ async function getSourceLocationPrefix(codeql: string) {
...
+    queries,
+  };
+}
+
+async function querySupportsSarif(
+  codeql: string,
+  queryPath: string,
+  bqrsInfo: BQRSInfo,
+): Promise<boolean> {
+  const compatibleQueryKinds = bqrsInfo.compatibleQueryKinds;
```
robertbrignull, line 349: (Open question)

The plumbing of all this is a bit odd. We call `codeql bqrs info` on every BQRS file earlier and store it so we can access it here and also call `codeql resolve metadata`. My brain is just saying it seems a bit off, but I can't think of anything better.

Is there any risk that we'll want to use the metadata somewhere else as well? Should we be storing it in `QueryPackRunResults` or otherwise computing it once and storing it, rather than computing it here?

src/codeql.ts
```diff
@@ -256,40 +399,21 @@ export function getSarifOutputType(
...
-    ),
     "--sarif-add-snippets",
     "--no-group-results",
-    // Hard-coded the source archive as src.zip inside the database, since that's
-    // where the CLI puts it. If this changes, we need to update this path.
-    `--source-archive=${databaseName}/src.zip`,
-    `--source-location-prefix=${sourceLocationPrefix}`,
-    bqrs,
+    databaseName,
+    queryPackName,
```
robertbrignull, line 416: Why are we using query pack name here rather than the query pack path? We seem to be using this all over the place, and sometimes (e.g. in `getRemoteQueryPackQueries`) we path both.

This isn't really being changed by this PR since the `getQueryPackName` function isn't being changed and the value was already being used in multiple places, but still it could be worth investigating what this value is and do we want to be using it over the path?

src/codeql.ts
```diff
@@ -256,40 +399,21 @@ export function getSarifOutputType(
...
 ): Promise<Sarif> {
-  const {
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars -- we are explicitly excluding this since it's calculated separately
-    kind,
-    id: queryId,
-    ...passthroughQueryMetadata
-  } = queryMetadata;
-
   const sarifFile = path.join("results", "results.sarif");
   await exec(codeql, [
```
robertbrignull, line 408: Random thought: how does this CodeQL CLI operation know to only generate SARIF for the correct queries? Does it automatically exclude non-problem queries? What happens if we call this when there aren't any problem queries in the pack?

My thought is, do we need to do the metadata and compatible-query-kinds parsing ourselves? Could we call this command unconditionally and use the output from the this command to determine if there were any problem queries or not?

src/codeql.ts
```diff
@@ -76,51 +77,97 @@ export async function runQuery(
...
+}
+
+async function moveBqrsFileToResultsDir(
+  resultsBasePath: string,
+  relativeBqrsFilePath: string,
+): Promise<string> {
+  const oldPath = path.join(resultsBasePath, relativeBqrsFilePath);
+  const newPath = path.join("results", relativeBqrsFilePath);
+
+  if (!fs.existsSync(path.dirname(newPath))) {
```
robertbrignull, line 163: This could have a high potential for a concurrency bug where the directory is created after we check for it. We'll be performing this operation for many files at once in parallel.

Do we gain much performance by doing all the calls to `moveBqrsFileToResultsDir` at once or would it be safer to make each call synchronous, i.e. use a for loop instead of `Promise.all`?

src/codeql.ts
```diff
@@ -76,51 +77,97 @@ export async function runQuery(
...
     databaseSHA,
+    databaseName,
     sourceLocationPrefix,
-    bqrsFilePath,
+    bqrsFilePaths,
     sarifFilePath,
   };
 }
 
+async function adjustBqrsFiles(
```
robertbrignull, line 127: Do we need to move the physical BQRS files around on disk? When we create the results artifact we decide which file goes at which path in the zip.

robertbrignull (commented):
Apologies this isn't a particular focused review. There's a lot going on this PR and a few places where I'm not sure what the best approach is, or how exactly the CodeQL CLI behaves. I've left a few comments that are quite open. We don't necessarily have to address everything in this PR, but I'd like to talk about everything at least.

Also, I tried it out and it failed. The failure seems to be related to changes in this PR rather than something wrong with my local setup: https://github.com/dsp-testing/qc-controller/actions/runs/7474330671/job/20340259609

src/codeql.ts
```diff
@@ -230,6 +277,102 @@ async function getSourceLocationPrefix(codeql: string) {
...
+}
+
+async function queryPackSupportsSarif(
+  codeql: string,
+  queriesResultInfo: QueryPackRunResults,
+) {
+  // Some queries in the pack must support SARIF in order
+  // for the query pack to support SARIF.
+  return (
+    await Promise.all(
```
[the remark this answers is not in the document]
charisk: I've done a few tests and it didn't seem to cause any issues, although I must admit I'm also a bit worried about it. Perhaps we should just turn it into a synchronous call and only look at parallelizing if it's too slow?

I've also looked for a different command to help us but didn't find anything.

src/codeql.ts
charisk: There is a [`RunQueryResult`](https://github.com/github/codeql-variant-analysis-action/blob/cc327c42612adbc12416fd3bebf0c312caaeef44/src/codeql.ts#L13) so I wanted to differentiate from that. In general naming in this part of the codebase seems to have room for improvement, but I didn't want to make too many changes.

src/codeql.ts
charisk: Agree it's a bit off and it's caused me some confusion. However, the changes are following the existing flow. I'd be happy to change it (and I'm glad you're equally finding it odd!), but I think it would be good to do that separately (in another PR/issue).

src/codeql.ts
charisk: That's a good point, I'll check. I had assumed it would fail but perhaps not.

src/codeql.ts
charisk: query.ts expects things to be in a `/results` dir so it seems a bit leaky to move that knowledge there.

src/codeql.ts
charisk: Shall we deal with this as a follow up? I don't really know why we use pack name in some places and pack path in others.

charisk:
> Also, I tried it out and it failed. The failure seems to be related to changes in this PR rather than something wrong with my local setup: https://github.com/dsp-testing/qc-controller/actions/runs/7474330671/job/20340259609

Can you tell me a bit more about what you ran so I can repro? Running a standard MRVA worked for me when I tested it.

robertbrignull:
> > Also, I tried it out and it failed. The failure seems to be related to changes in this PR rather than something wrong with my local setup: https://github.com/dsp-testing/qc-controller/actions/runs/7474330671/job/20340259609
> 
> Can you tell me a bit more about what you ran so I can repro? Running a standard MRVA worked for me when I tested it.

What I did:
- Language is Java
- Query is the generic "empty blocks" example query
- Repo is `jhy/jsoup`
- And `"codeQL.variantAnalysis.actionBranch": "charisk/add-multi-query-support"` set in my config

Here's another failing run: https://github.com/dsp-testing/qc-controller/actions/runs/7475054635

If I set the action branch back to `main` then it succeeds: https://github.com/dsp-testing/qc-controller/actions/runs/7475077387

src/codeql.ts
koesie10: I think you can just call `fs.promises.mkdir` unconditionally since the `recursive` option is set to `true`. According to [the documentation](https://nodejs.org/docs/latest-v20.x/api/fs.html#fspromisesmkdirpath-options) "calling `fsPromises.mkdir()` when `path` is a directory that exists results in a rejection only when `recursive` is `false`."

src/codeql.ts
robertbrignull: It might fail, but if it fails saying "no suitable queries found" then that could still be acceptable. 🤷🏼

charisk: It looks like a SARIF files is created, but with 0 results (https://github.com/dsp-testing/qc-controller/actions/runs/7475835885/job/20344865094#step:10:1447). I don't think that's what we want though.

src/codeql.ts
robertbrignull: This repo is small enough that I think it's ok to have a slightly-more-complex interface between functions/files if it avoids us making the code more complex and having to do computational work to fit those interfaces.

Although, perhaps we can make the interface even more clear and also avoid copying the files around. Currently the unspoken assumption is that the BQRS files are contained in the `/results` dir, but how about if `RunQueryResult` included a `bqrsFilePathBase` (or whatever name) that indicates what all the BQRS files should be considered "relative to". That way `query.ts` doesn't have to assume it's the `/results` directory and we can keep the BQRS files where they are.

charisk: Done!

src/codeql.ts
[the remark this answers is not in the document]
charisk: I've updated to a more serial implementation.

robertbrignull (approved):
LGTM to merge

I've tested it manually and it's still able to successfully run a single query.

I think all of the major points from my earlier review have been addressed now. I'm happy for all of the other points to be looked at later in issues or followup PRs.

src/codeql.ts
charisk: Created internal issue to cover this.

src/codeql.ts
charisk: Created internal issue to cover this.

charisk:
I've ran all the required test cases from the VS Code extension test plan and did some more manual exploratory testing, against this branch. Existing functionality seems to be working as before so I'll go ahead and merge.

PR #972, opened 2024-01-22
Rename queryPack => queryPackPath

robertbrignull:
Everywhere we currently have a variable named `queryPack` I've renamed it to `queryPackPath`, as in all cases this is what the variable contained. I did also consider `queryPackDir`, but we have a number of existing `queryPackPath` variables in this repo so I decided to stay consistent with those and not change too much in this PR.


charisk (approved):
LGTM

PR #974, opened 2024-01-24
Add octokit throttling behind env var

jorgectf:
This pull request primarily focuses on enhancing the rate limit handling in the `gh-api-client.ts` file by integrating the `@octokit/plugin-throttling` library. It also includes a minor change to the `variant-analysis-workflow.yml` file to set an environment variable.

Rate limit handling enhancement:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R37): Added `@octokit/plugin-throttling` to the list of dependencies.
* [`src/gh-api-client.ts`](diffhunk://#diff-edee0dd739a0d285c7e996010b790ae84a434a4629984aed47ca48b1aad6c135R4): Imported the `throttling` function from `@octokit/plugin-throttling`.
* [`src/gh-api-client.ts`](diffhunk://#diff-edee0dd739a0d285c7e996010b790ae84a434a4629984aed47ca48b1aad6c135L15-R48): Refactored the `getOctokitRequestInterface` function to `getOctokit` and integrated the throttling plugin into the Octokit instance. This new setup will handle rate limits by waiting and retrying requests when limits are reached.
* [`src/gh-api-client.ts`](diffhunk://#diff-edee0dd739a0d285c7e996010b790ae84a434a4629984aed47ca48b1aad6c135L166-R190): Updated the `updateVariantAnalysisStatus`, `updateVariantAnalysisStatuses`, and `getPolicyForRepoArtifact` functions to use the new `getOctokit` function. [[1]](diffhunk://#diff-edee0dd739a0d285c7e996010b790ae84a434a4629984aed47ca48b1aad6c135L166-R190) [[2]](diffhunk://#diff-edee0dd739a0d285c7e996010b790ae84a434a4629984aed47ca48b1aad6c135L184-R208) [[3]](diffhunk://#diff-edee0dd739a0d285c7e996010b790ae84a434a4629984aed47ca48b1aad6c135L208-R232)

Environment variable setting:

* [`variant-analysis-workflow.yml`](diffhunk://#diff-6a5e187c4aa666a1c04567c14c873a01db49511f8756a04ac6dd34ee7520ec9dR7): Added the `CODEQL_VARIANT_ANALYSIS_ACTION_WAIT_ON_RATE_LIMIT` environment variable, which enables or disables the rate limit handling based on the `MRVA_WAIT_ON_RATE_LIMIT` variable.

### DIFF
--- a/src/gh-api-client.ts
+++ b/src/gh-api-client.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { Octokit } from "@octokit/action";
 import { retry } from "@octokit/plugin-retry";
+import { throttling } from "@octokit/plugin-throttling";
 import {
   EndpointOptions,
   RequestError,
@@ -12,20 +13,39 @@ import { validateObject } from "./json-validation";
 
 export const userAgent = "GitHub multi-repository variant analysis action";
 
-function getOctokitRequestInterface(): RequestInterface {
-  const octokit = new Octokit({ userAgent, retry });
-
-  const signedAuthToken = getSignedAuthToken();
-  return octokit.request.defaults({
-    request: {
-      hook: (request: RequestInterface, options: EndpointOptions) => {
-        if (options.headers) {
-          options.headers.authorization = `RemoteAuth ${signedAuthToken}`;
-        }
-        return request(options);
+function getOctokit(): Octokit {
+  const throttlingOctokit = Octokit.plugin(throttling);
+  const octokit = new throttlingOctokit({
+    userAgent,
+    retry,
+    authStrategy: () => {
+      return {
+        hook: (request: RequestInterface, options: EndpointOptions) => {
+          if (options.headers) {
+            options.headers.authorization = `RemoteAuth ${getSignedAuthToken()}`;
+          }
+          return request(options);
+        },
+      };
+    },
+    throttle: {
+      enabled: !!process.env.CODEQL_VARIANT_ANALYSIS_ACTION_WAIT_ON_RATE_LIMIT,
+      onRateLimit: (retryAfter, options) => {
+        console.log(
+          `Rate limit exhausted for request ${options.method} ${options.url}, retrying after ${retryAfter} seconds`,
+        );
+        return true;
+      },
+      onSecondaryRateLimit: (retryAfter, options) => {
+        console.log(
+          `Secondary rate limit triggered for request ${options.method} ${options.url}, retrying after ${retryAfter} seconds`,
+        );
+        return true;
       },
     },
   });
+
+  return octokit;
 }
 
 export interface Policy {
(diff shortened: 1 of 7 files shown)


src/gh-api-client.ts
```diff
@@ -12,20 +13,39 @@ import { validateObject } from "./json-validation";
...
+        hook: (request: RequestInterface, options: EndpointOptions) => {
+          if (options.headers) {
+            options.headers.authorization = `RemoteAuth ${getSignedAuthToken()}`;
+          }
+          return request(options);
+        },
+      };
+    },
+    throttle: {
+      enabled: !!process.env.CODEQL_VARIANT_ANALYSIS_ACTION_WAIT_ON_RATE_LIMIT,
```
robertbrignull, line 32: Does this correctly handle a "false" value?

I believe `process.env.CODEQL_VARIANT_ANALYSIS_ACTION_WAIT_ON_RATE_LIMIT` will always be a string value, but the value may be the strings "true" or "false". So then using `!!` on this will always yield `true`, because javascript just looks at whether the string is non-empty or not.

That said, do we need this functionality to be gated behind a feature flag or is it something we should enable for all users? I can't see any downsides of it or why a user would want to turn it off. 🤷🏼

src/gh-api-client.ts
```diff
@@ -12,20 +13,39 @@ import { validateObject } from "./json-validation";
...
-
-  const signedAuthToken = getSignedAuthToken();
-  return octokit.request.defaults({
-    request: {
-      hook: (request: RequestInterface, options: EndpointOptions) => {
-        if (options.headers) {
-          options.headers.authorization = `RemoteAuth ${signedAuthToken}`;
-        }
-        return request(options);
+function getOctokit(): Octokit {
```
robertbrignull, line 16: Thanks for cleaning up this interface! ❤️ 

The naming and returning that `RequestInterface` type was a bit awkward, but at the time I couldn't find a better way of doing things. What you've got here looks much better!

robertbrignull (commented):
Adding in this functionality sounds good to me. So long as manual and automated tests pass then I'm happy 👍🏼

src/gh-api-client.ts
jorgectf: That's a great catch! IIRC from the testing I did, the absence of the variable repository made `enabled: false`, but I agree it's not ideal.

> do we need this functionality to be gated behind a feature flag

I'm guessing we do, as private repositories would get their Actions usage increased a lot ($$$) in case of a rate limit triggered, but I'm happy to remove the "gate".

src/gh-api-client.ts
jorgectf: Thank you! I spent a bit figuring out how to simulate adding the hook without returning `octokit.request.defaults`, which does not carry the rest of the features, but reversing a builtin [auth strategy](https://github.com/octokit/auth-token.js/blob/main/src/index.ts#L27) I saw it was almost the same implementation 🪄

src/gh-api-client.ts
jorgectf: > That makes sense, but is the variable ever absent, because we do ${{ (vars.MRVA_WAIT_ON_RATE_LIMIT && true) || false }}? In the false case what value does the env var get inside nodejs?

Oops, I just re-tried and it works as you say, so https://github.com/github/codeql-variant-analysis-action/pull/974/commits/162ff3bdf1179e31ee868fc83d71f8506e57d630 makes the default value a blank string to fix the discrepancy.

> Would it make sense to add a limit so it only retries a bounded number of times? That way it wouldn't increase costs by very much.

That would work, but still increase the use a lot and not work for when the rate limit is triggered more than once at a time.

> Out of interest, who are the intended users right now? (You're welcome to message us on slack or in an internal issue if you don't want to say here on this public PR).

Me! We are doing heavy use of MRVA lately and this helps me launch queries without worrying about them being canceled.

charisk (approved):
LGTM

PR #982, opened 2024-02-05
Read problems result set if '#select' result set doesn't exist

robertbrignull:
MRVA fails when running certain queries, such as [`IncompleteHostnameRegExp.ql`](https://github.com/github/codeql/blob/main/python/ql/src/Security/CWE-020/IncompleteHostnameRegExp.ql). This is because the resulting BQRS file doesn't contain a `#select` results set but instead contains one called `problems`.

This PR allows us to use either results set. In my manual searches, every query I've looked at has precisely one of these two.

I've tested this PR by confirming that `IncompleteHostnameRegExp.ql` fails when using the `main` branch of this repo, but passes when using this branch. I then also tested this branch with another `problem` query, a `path-problem` query, a `table` query, and also tried running against the published pack for python.

---

I've also opened an internal issue to look at reducing/removing our calls to `codeql bqrs info`, though I'm not sure we'll be able to remove them entirely, so we'll still want this fix in there.

### DIFF
--- a/src/codeql.ts
+++ b/src/codeql.ts
@@ -450,16 +450,27 @@ export function getSarifResultCount(sarif: Sarif): number {
 }
 
 /**
+ * Names of result sets that can be considered the "default" result set
+ * and should be used when calculating number of results and when showing
+ * results to users.
+ */
+const KNOWN_RESULT_SET_NAMES: string[] = ["#select", "problems"];
+
+/**
  * Gets the number of results in the given BQRS data.
  */
-function getBqrsResultCount(bqrsInfo: BQRSInfo): number {
-  const selectResultSet = bqrsInfo.resultSets.find(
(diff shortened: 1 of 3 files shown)


src/codeql.ts
```diff
@@ -449,17 +449,28 @@ export function getSarifResultCount(sarif: Sarif): number {
   return count;
 }
 
+/**
+ * Names of result sets that can be considered the "default" result set
+ * and should be used when calculating number of results and when showing
+ * results to users.
+ */
+const KNOWN_RESULT_SET_NAMES: string[] = ["#select", "problems"];
```
charisk, line 457: This list is also setting the order by which we'll look for result sets. It would be great if the name could make that obvious (although I'm struggling to think of a good suggestion)

charisk (approved):
LGTM!

PR #984, opened 2024-02-09
Retry request timeout errors

koesie10:
When we get a request timeout while uploading an artifact, a [request timeout error is created by the Actions HTTP client](https://github.com/actions/toolkit/blob/415c42d27ca2a24f3801dd9406344aaea00b7866/packages/http-client/src/index.ts#L535). We want to retry on this error since this is very likely to be transient. Unfortunately, it's not a custom error which we can check against, so we need to match against the message.


charisk (approved):
Thanks. As you said, it's not great that we have to hardcode a part of the message but the text we use is so specific and clear that I think it's okay.

PR #988, opened 2024-02-12
Allow specifying a custom CLI path in tests

koesie10:
This allows you to run tests when you don't have the CLI in your `PATH`. For example, if you're using [gh codeql](https://github.com/github/gh-codeql), you can set `CODEQL_BIN_PATH=~/.local/share/gh/extensions/gh-codeql/gh-codeql`.


robertbrignull (approved):
Thanks for that. It'll make it much easier to test.

PR #989, opened 2024-02-12
Create temporary directories in OS temp dir

koesie10:
Temporary directories were being created in the working directory rather than in the system temporary directory. `mkdtemp` only generates a random directory name, but doesn't actually use the system temporary directory. This fixes it by adding a call to `os.tmpdir()` as shown [in the examples](https://nodejs.org/docs/latest-v18.x/api/fs.html#fsmkdtempprefix-options-callback).

It also unifies the creation of temporary directories so that we don't need to make all of these calls for every test.

### DIFF
--- a/src/codeql.test.ts
+++ b/src/codeql.test.ts
@@ -1,4 +1,5 @@
 import fs from "fs";
+import { tmpdir } from "os";
 import path from "path";
 
 import { exec } from "@actions/exec";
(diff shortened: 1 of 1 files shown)


src/codeql.test.ts
```diff
@@ -20,14 +21,18 @@ import {
   getQueryPackInfo,
 } from "./codeql";
 
-const test = anyTest as TestFn<{ db: string; tmpDir: string }>;
+const test = anyTest as TestFn<{
```
robertbrignull, line 24: This will conflict with https://github.com/github/codeql-variant-analysis-action/pull/988 but should be easy to fix up

robertbrignull (approved):
Thanks for spotting that and clearing it up

PR #990, opened 2024-02-12
Use the CodeQL CLI server for running CodeQL commands

koesie10:
This PR adds the ability to use the CodeQL CLI server for running commands. This results in better performance of individual command invocations since there's less JVM overhead. On a test I ran, the overhead (defined as time not spent running `codeql database run-queries`) went down from 20% to 3-5%.

This is currently guarded by the `MRVA_USE_CLI_SERVER` repo Actions variable. We can probably remove this in a follow-up PR once this has been sufficiently tested.

### DIFF
--- a/dist/query.js
+++ b/dist/query.js
@@ -75401,6 +75401,9 @@ var CodeqlCliServer = class {
       }
     });
   }
+  shutdown() {
+    this.killProcessIfRunning();
+  }
   /**
    * Launch the cli server
    */
@@ -75526,6 +75529,7 @@ var CodeqlCliServer = class {
 };
 
 // src/query.ts
+var shutdownHandlers = [];
 async function run() {
   const controllerRepoId = getControllerRepoId();
   const queryPackUrl = (0, import_core3.getInput)("query_pack_url", { required: true });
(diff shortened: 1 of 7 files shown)


robertbrignull:
Interesting. We might have to think hard about whether it's ok to use the CLI server in this repo, since it's such an internal command. However, I'm coming to the conclusion that we may as well, since it's used in vscode-codeql which we also own and is more important than this repo. The reason not to use the CLI server here is that it increases maintenance burden on us and the CodeQL team if it ever breaks, but I think the burden will be minimal. This repo isn't particularly important and is easy to change and fix bugs in, and in the worst case we can always remove use of the CLI server and go back to the regular CLI.

So overall if we accept the risk then I think this is fine, because it's us and we're very close to the CLI team and to the vscode-codeql repo so we'll hear about any changes to the CLI server.

robertbrignull (approved):
I've had a read through of the code and it looks fine. I recognise a lot of the stuff for the CLI server is copied from the vscode extension, and that's fine.

I've tried this out with the variable enabled and disabled, and I think it's working correctly.

PR #995, opened 2024-03-01
Add support for alert and path-alert aliases

koesie10:
This adds support for the `alert` and `path-alert` aliases to treat them as `problem` and `path-problem`, respectively.


norascheuch (approved):
🚢

PR #1006, opened 2024-03-18
Group some npm dependency updates

koesie10:
See https://github.com/github/vscode-codeql/pull/3477 for the same changes in the extension


norascheuch (approved):
Nice!

PR #1028, opened 2024-04-26
Add support for GHEC-DR

koesie10:
This adds support for GHEC-DR. It replaces hardcoded `github.com` URLs by [the environment variables available in Actions](https://docs.github.com/en/actions/learn-github-actions/variables) with a fallback to the current value. This matches [what `@octokit/action` does](https://github.com/octokit/action.js/blob/aee7aaedceedcf7cd64db65cc9aff03bf95f3f25/src/index.ts#L56C10-L56C67). Since `@octokit/action` already does this, we don't need to make any changes to Octokit constructors and this is handled automatically.


charisk (approved):
Nice!

PR #1037, opened 2024-05-08
Switches from the Ava to Jest testing framework

robertbrignull:
I followed the guide for getting started from the Jest docs, including using `npm init jest` to generate a config file. I also followed the example of the `vscode-codeql` repo by using `ts-jest`.

It's a pretty easy conversion. Hopefully if you ignore whitespace changes you can see the changes are purely different function names or ways of doing before/after hooks.

If you ignore the `package-lock.json` and `jest.config.ts` files which are both generated, then it's no change in the number of lines of code.

I had to set custom timeouts on a few tests, but I think this is reasonable as they're doing hefty work like building CodeQL databases. I just picked a number out of the air for these timeout values, so feel free to suggest different values.


src/codeql.test.ts
```diff
@@ -21,376 +20,414 @@ import {
...
-});
-
-test("getting query pack info", async (t) => {
-  const queryPackInfo = await getQueryPackInfo(
-    t.context.codeql,
-    "testdata/test_pack",
+describe("codeql", () => {
+  const codeql: CodeqlCli = new BaseCodeqlCli(
+    process.env.CODEQL_BIN_PATH || "codeql",
   );
```
koesie10, line 26: Should we be consistent in when we create this object? In the `codeql-cli.test.ts` we create this in the `beforeAll` hook, while here we create it in the `describe` block. I have a preference for doing this in the `beforeAll` hook, but I'll leave it up to you which option you think is best.

koesie10 (approved):
LGTM, just a suggestion regarding an inconsistency between some test files (and there seems to be a failing lint because the `jest.config.ts` file is not included in the TS project).

src/codeql.test.ts
robertbrignull: I've moved this to be in `beforeAll` so we're more consistent and avoiding re-using objects between tests.

robertbrignull:
> there seems to be a failing lint because the `jest.config.ts` file is not included in the TS project

I looked into this and tried including it in the project. Sadly it fails `filenames/match-regex` so we'd have to disable that rule on this file, but it's doable.

However the solution I've pushed is to exclude `jest.config.ts` from linting. I think this is fine because that file is basically a single object with options and we don't gain much from linting it.

PR #1083, opened 2024-10-10
Update variant-analysis-workflow.yml

CallMeGreg:
This pull request includes changes to the `variant-analysis-workflow.yml` file to make the workflow runner dynamic across all jobs, instead of just one of the jobs, for organizations that only use self-hosted runners.

### DIFF
--- a/variant-analysis-workflow.yml
+++ b/variant-analysis-workflow.yml
@@ -10,7 +10,7 @@ on: dynamic
 
 jobs:
   setup:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


robertbrignull:
This change looks good to me, however this isn't the right file to change anymore and the change needs to happen in a private repo too. This workflow on the default branch isn't actually used anymore and I'll be looking at deleting it soon to avoid confusion.

I'll make the necessary change in the right place and then close this PR once I've done that.

PR #1089, opened 2024-10-17
Delete variant-analysis-workflow.yml

robertbrignull:
This workflow file is no longer used in production. The workflow is now baked into github instead of being read from here. We've also now updated the documentation around testing changes to not rely on this file being here or being up to date on the default branch.


koesie10:
@robertbrignull I think the integration tests don't work using these changes, it seems like the Actions workflow run ID is `null`: https://github.com/github/codeql-variant-analysis-action/actions/runs/11381738118/job/31663793721?pr=1089#step:3:124

robertbrignull:
Hmm, I hadn't spotted that. In the integration tests we pass `action_repo_ref`, meaning the integration tests do currently require that the workflow file is present on the default branch.

Guess this will have to wait a little longer. I'll reopen the internal issue and post my thoughts there.

PR #1218, opened 2025-06-10
Upgrade to eslint version 9

robertbrignull:
This PR upgrades ESLint to version 9.x.x. The conversion was quite involved, and there is a conversion tool but it found it was quite confusing and didn't work too well so I had to do it manually. It's been a bit of a struggle and an adventure but this is what I've ended up with. As far as I can tell it's all working 🤞🏼 

This PR supersedes https://github.com/github/codeql-variant-analysis-action/pull/1212 and https://github.com/github/codeql-variant-analysis-action/pull/1185.

There are a couple of small changes I made during the migration:
- I removed our direct dependency on `eslint-plugin-import` because it's already included in `eslint-plugin-github` (and was causing errors trying to redeclare it in the config file). As far as I can tell this hasn't changed any behaviour.
- I removed `eslint-plugin-filenames` because it's quite old and even using `compat` it wasn't working. I just wasn't sure it was important enough to continue messing around with so I removed it, but please say if you think that was a mistake.


koesie10 (approved):
LGTM, and the config looks like it matches.

PR #1456, opened 2026-08-04
Align ESLint dependencies on v10

henrymercer:
## Summary

- upgrade `eslint` from 9.39.5 to 10.8.0
- align the declared `@typescript-eslint/parser` minimum with the already-resolved ESLint 10-compatible 8.65.0 release
- regenerate `package-lock.json` from a clean install

## Context

The manifest previously combined `@eslint/js@^10.0.1`, whose peer dependency requires ESLint 10, with `eslint@^9.39.5`. The committed lockfile preserved that mismatched graph, so `npm ci` worked, but npm could not regenerate the lockfile from the manifest.

ESLint 10 was previously postponed because `eslint-plugin-github` only supported ESLint 9. That blocker is resolved: `eslint-plugin-github` supports ESLint 10 as of 6.1.0, and this repository already uses `^6.1.1`.

Dependabot is currently configured to ignore the ESLint 10.x major after the earlier postponement. Someone may want to lift that ignore after this merges so future ESLint 10.x updates flow normally; this PR intentionally does not change Dependabot configuration.

## Validation

- `npm ci`
- `npm run lint`
- `.github/workflows/script/check-js.sh`
- `.github/workflows/script/check-json-schemas.sh`
- `npm run find-deadcode`
- `npm test`


mbg (approved):

oscarsj:
@henrymercer what happened here?

henrymercer:
`eslint-plugin-github` supports eslint v10 on its own, but it also pulls in `eslint-plugin-import@2.32.0` and `eslint-plugin-jsx-a11y@6.10.2` which don't support eslint v10.  The linting passes, but it's not supported and we'll get peer dependency warnings in our builds, so I suggest we stick with v9 for now.

PR #1457, opened 2026-08-04
Dependabot: Change `npm` groupings to prioritise grouping minor/patch version bumps

mbg:
Follow-up change to #1413 since we don't need a separate group for minor/patch-level `octokit` updates. One group for all minor/patch-level updates is enough.


henrymercer (commented):
I think this will create separate PRs for each major/minor/patch octokit bump, and instead we should include `update-types: ["major"]`.

PR #1459, opened 2026-08-04
Remove five unused `@babel/*` devDependencies

henrymercer:
The five `@babel/*` packages in `package.json` are entirely unused by this repository. Removing them resolves a Dependabot deadlock permanently, rather than moving it forward one version.

## Why these packages are unused

- **Build** is esbuild. `build.mjs` imports `esbuild` and nothing else; esbuild handles TypeScript and JSX itself, with no Babel transform hook.
- **Tests** are Vitest, which transforms via esbuild/Vite and handles ESM natively.
- **No Babel configuration exists** anywhere in the repo — no `babel.config.*`, no `.babelrc*`, and no `vitest.config.*` or `jest.config.*` that could reference one.
- **No Babel imports or API usage** in any source, script or config file.
- **Linting does not involve Babel.** `eslint.config.mjs` parses with `@typescript-eslint/parser`; `@babel/eslint-parser` is not installed, and none of the five is a lint tool.
- **Lockfile reverse-edge analysis** shows no used package depends on any of the five. Their only inbound edges come from the root manifest and from Babel's own packages; `babel-plugin-polyfill-*` is reachable only via `@babel/preset-env`, and the many `@babel/core` peer edges all originate from plugins that exist solely because those presets pull them in. Neither `eslint-plugin-github` nor `vitest` has a Babel dependency.

## Impact

Removing them drops **129 of 682 locked packages** — roughly 19% of the dependency tree — and 2,011 lines from `package-lock.json`, with nothing added. That is 111 Babel-named packages (107 `@babel/*` plus 4 `babel-plugin-polyfill-*`) and 18 transitive dependencies reachable only through them, such as `core-js-compat`, `regexpu-core` and the `unicode-*-ecmascript` helpers.

`dist/` is unchanged, as expected for a dev-only removal: nothing Babel-related was ever bundled. The three `dist/*.js` matches for "Babel" are esbuild's own generated `__toESM` interop comment explaining the `__esModule` check, not bundled Babel code.

## Validation

`npm ci`, `npm run lint`, `check-js`, `check-json-schemas`, `npm run find-deadcode` and the full Vitest suite all pass.

## Notes

- Stacked on #1458, which aligns `@eslint/js` back to `^9.39.5` to match `eslint`. That pre-existing mismatch made `npm install` fail with `ERESOLVE`, so `package-lock.json` could not be regenerated at all until it landed. Please merge #1458 first.


mbg (approved):
Nice

